PR #329, opened 2015-01-04
check for mono and configure appropriately

friism:
This is just copied from the non-SPA project


Praburaj:
We don't do direct merges to master. We can take the PR you sent to the dev branch forward.

friism:
@Praburaj Understood, thanks

PR #332, opened 2015-01-06
Updating path to kvm

suhasj:
@Praburaj 

This is blocking build on CI. Updating the path to point to the right kvm.sh file


Praburaj:
:shipit:

PR #342, opened 2015-01-15
Increase wait time for server to start

suhasj:
@Praburaj

### DIFF
--- a/test/E2ETests/Common/DeploymentUtility.cs
+++ b/test/E2ETests/Common/DeploymentUtility.cs
@@ -201,7 +201,7 @@ namespace E2ETests
 
             var hostProcess = Process.Start(startInfo);
             logger.WriteInformation("Started {0}. Process Id : {1}", hostProcess.MainModule.FileName, hostProcess.Id);
-            Thread.Sleep(15 * 1000);
+            Thread.Sleep(25 * 1000);
 
             //Clear the appbase so that it does not create issues with successive runs
             Environment.SetEnvironmentVariable("KRE_APPBASE", string.Empty);


test/E2ETests/Common/DeploymentUtility.cs
```diff
@@ -201,7 +201,7 @@ private static Process StartMonoHost(StartParameters startParameters, ILogger lo
 
             var hostProcess = Process.Start(startInfo);
             logger.WriteInformation("Started {0}. Process Id : {1}", hostProcess.MainModule.FileName, hostProcess.Id);
-            Thread.Sleep(15 * 1000);
+            Thread.Sleep(25 * 1000);
```
Eilon: Is there no way to "know" whether the target process is done doing its thing? Can the target process somehow notify this process that it's done? Maybe drop a file somewhere as a signal? Or something?

suhasj: We are basically waiting for Kestrel to start. I can open a bug on Kestrel to provide some notification on start

Praburaj: On windows we wait for the local db file to be created. On mono since we use inmemory store we can't apply the same logic. For now looks good to me :shipit:

Eilon: You can make requests to the server just waiting for it to response, that's a pretty common technique. This test will now *always* take 25 seconds even if the server took only 1ms to start.

Praburaj:
Alternatively we could use the retryhelper. Currently it takes retryCount, we can make it to take in a wait time as well - https://github.com/aspnet/MusicStore/blob/dev/test/E2ETests/Common/Helpers.cs#L19.

Eilon:
That could indeed be better.

PR #350, opened 2015-01-16
Fix music store policy definitions

HaoK:
Didn't setup policy correctly

cc @Praburaj


HaoK:
Merged 464438862a41432b168b215df16f0a47dc6ebba5

PR #365, opened 2015-01-21
React to default packages path change in runtime

ChengTian:
Changing default packages path to `%USERPROFILE%/.kpm/packages` to `%USERPROFILE%/.dotnet/packages`

@davidfowl @Praburaj


Praburaj:
Rebase and :shipit:

PR #369, opened 2015-01-23
Removed BuddyType attribute and added ModelMetadatType 

kirthik:
Removed BuddyType attribute and added ModelMetadatType instead. Added AutoMapper for Desktop scenario. I am still using SimpleMapper for Core scenario.

@DamianEdwards, @yishaigalatzer - Could you please take a look?


src/MusicStore.Spa/Apis/AlbumsApiController.cs
```diff
@@ -25,14 +28,20 @@ public AlbumsApiController(MusicStoreContext storeContext)
...
+            //.Include(a => a.Artist)
+#if ASPNET50
+            await albums.ToPagedListAsync(page, pageSize, sortBy,
+                    a => a.Title,                                    // sortExpression
+                    SortDirection.Ascending,                         // defaultSortDirection
+                    a => Mapper.Map(a, new AlbumResultDto()));       // selector
+#else
+            await albums.ToPagedListAsync(page, pageSize, sortBy,
                     a => a.Title,                                    // sortExpression
                     SortDirection.Ascending,                         // defaultSortDirection
```
yishaigalatzer: nit: you can just `ifdef` the relevant line

yishaigalatzer:
looks much cleaner to me.

@DamianEdwards we also finished the action result pattern that we thought made sense, should we follow up with a second PR to remove ApiResult?

src/MusicStore.Spa/Apis/AlbumsApiController.cs
```diff
@@ -99,7 +122,11 @@ public AlbumsApiController(MusicStoreContext storeContext)
 
             // Save the changes to the DB
             var dbAlbum = new Album();
+#if ASPNET50
+            await _storeContext.Albums.AddAsync(Mapper.Map(album, dbAlbum));
```
yishaigalatzer: Wrap the mappers into a single mapper abstraction

src/MusicStore.Spa/Infrastructure/SimpleMapper.cs
```diff
@@ -14,8 +14,12 @@ public class SimpleMapper
 
         public static TDest Map<TSource, TDest>(TSource source, TDest dest)
         {
+#if ASPNET50
+            return AutoMapper.Mapper.Map(source, dest);
+#else
             var map = (Func<TSource, TDest, TDest>)_mapCache.GetOrAdd(Tuple.Create(typeof(TSource), typeof(TDest)), _ => MakeMapMethod<TSource, TDest>());
             return map(source, dest);
+#endif
```
kirthik: @yishaigalatzer - Moved AutoMapper mapping to SimpleMapper Map method. Does this look better?

yishaigalatzer: Much better, even better would be to wrap in a different interface 

-----Original Message-----

rynowak:
@kirthik is there a reason this is still sitting here?

kirthik:
Oops no, lost track of it. I had to get sign off from @DamianEdwards and @yishaigalatzer. I will get it in ASAP.

rynowak:
awesome! this is going to unblock some of the stuff that I need to do here

kirthik:
If code looks fine to you, could you please sign off?

DamianEdwards:
@kirthik It seems you've deleted `SimpleMapper` completely? If so, what are we using in .NET Core scenario now?

yishaigalatzer:
Automapper now supports coreclr

DamianEdwards:
:shipit:

kirthik:
cc @rynowak - This is in.

rynowak:
Awesome thanks!

PR #382, opened 2015-02-04
Updating MusicStore for better unit-testability

ianhong:
@rynowak, @yishaigalatzer, and @Praburaj 

1. Make CancellationToken for an action parameter so that users don't have to directly access Context.RequestAborted. In this way, users don't have to mock HttpContext for CancellationToken. 

2. Use the property DI instead of the constructor DI for SignInManager, UserManager, DbContext, IConnectionManager, IMemoryCache so that users can selectively mock those objects.

Thanks,

Note:
1. I didn't change the original code styles. For example, I didn't do any extra indentation. 
2. I'll add the unit tests for each controller later

### DIFF
--- a/src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
+++ b/src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc;
 using Microsoft.AspNet.Mvc.Rendering;
@@ -17,26 +18,42 @@ namespace MusicStore.Areas.Admin.Controllers
     [Microsoft.AspNet.Mvc.Authorize("ManageStore")]
     public class StoreManagerController : Controller
     {
-        private readonly MusicStoreContext _dbContext;
-        private readonly IMemoryCache _cache;
+        private IConnectionManager _connectionManager;
         private IHubContext _announcementHub;
 
-        public StoreManagerController(
-            MusicStoreContext dbContext,
-            IConnectionManager connectionManager,
-            IMemoryCache memoryCache)
+        [Activate]
+        public MusicStoreContext DbContext
         {
-            _dbContext = dbContext;
-            _announcementHub = connectionManager.GetHubContext<AnnouncementHub>();
-            _cache = memoryCache;
+            get;
+            set;
+        }
+
+        [Activate]
+        public IMemoryCache MemoryCache
+        {
+            get;
+            set;
+        }
+
+        [Activate]
+        public IConnectionManager ConnectionManager
+        {
+            get
+            {
+                return _connectionManager;
+            }
+            set
+            {
+                _connectionManager = value;
+                _announcementHub = _connectionManager.GetHubContext<AnnouncementHub>();
+            }
         }
 
         //
         // GET: /StoreManager/
-
         public async Task<IActionResult> Index()
         {
-            var albums = await _dbContext.Albums
+            var albums = await DbContext.Albums
                 .Include(a => a.Genre)
                 .Include(a => a.Artist)
                 .ToListAsync();
@@ -46,18 +63,17 @@ namespace MusicStore.Areas.Admin.Controllers
 
         //
         // GET: /StoreManager/Details/5
-
         public async Task<IActionResult> Details(int id)
         {
             var cacheKey = GetCacheKey(id);
 
-            var album = await _cache.GetOrSet(cacheKey, async context =>
+            var album = await MemoryCache.GetOrSet(cacheKey, async context =>
             {
                 //Remove it from cache if not retrieved in last 10 minutes.
                 context.SetSlidingExpiration(TimeSpan.FromMinutes(10));
 
                 //If this returns null how do we prevent the cache to store this.
-                return await _dbContext.Albums
+                return await DbContext.Albums
                     .Where(a => a.AlbumId == id)
                     .Include(a => a.Artist)
                     .Include(a => a.Genre)
@@ -66,7 +82,7 @@ namespace MusicStore.Areas.Admin.Controllers
 
             if (album == null)
             {
-                _cache.Remove(cacheKey);
+                MemoryCache.Remove(cacheKey);
                 return View(album);
             }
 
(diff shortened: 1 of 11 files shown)


msotclas:
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla2.msopentech.com.
    </span>

TTYL, MSOTBOT;

src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
```diff
@@ -130,19 +146,19 @@ public IActionResult Create()
         // POST: /StoreManager/Edit/5
         [HttpPost]
         [ValidateAntiForgeryToken]
-        public async Task<IActionResult> Edit(Album album)
+        public async Task<IActionResult> Edit(CancellationToken requestAborted, Album album)
```
Praburaj: Can we make `CancellationToken` as the last parameter in all actions?

ianhong: Many actions use the default parameter value like string returnUrl = null). So, I can't use this the last parameter for all actions. For this action, I use this for the first parameter to keep consistency..but...I can make it as the last parameter for this particular action.

Praburaj: i wish the actual model binding object first in the parameter list. Probably make cancellation token last but one if there are default parameters?

ianhong: Sure. I'll make it as the last parameter in the case actions doesn't use the default parameter values..

pranavkm: Per this discussion, shouldn't Album appear first?

src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
```diff
@@ -150,23 +166,23 @@ public IActionResult Create()
...
-            var album = await _dbContext.Albums.Where(a => a.AlbumId == id).FirstOrDefaultAsync();
+            var album = await DbContext.Albums.Where(a => a.AlbumId == id).FirstOrDefaultAsync();
             return View(album);
         }
 
         //
         // POST: /StoreManager/RemoveAlbum/5
         [HttpPost, ActionName("RemoveAlbum")]
-        public async Task<IActionResult> RemoveAlbumConfirmed(int id)
+        public async Task<IActionResult> RemoveAlbumConfirmed(CancellationToken requestAborted, int id)
```
pranavkm: `int id` should appear first.

src/MusicStore/Controllers/ManageController.cs
```diff
@@ -13,19 +14,15 @@ namespace MusicStore.Controllers
...
-        public UserManager<ApplicationUser> UserManager { get; private set; }
-
-        public SignInManager<ApplicationUser> SignInManager { get; private set; }
+        [Activate]
+        public SignInManager<ApplicationUser> SignInManager { get; set; }
 
         //
         // GET: /Manage/Index
-        public async Task<ActionResult> Index(ManageMessageId? message = null)
+        public async Task<ActionResult> Index(CancellationToken requestAborted, ManageMessageId? message = null)
```
pranavkm: Maybe just call this `CancellationToken cancellationToken`? It's the common nomenclature for CancellationToken.

src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
ianhong: Yes..I haven't updated it yet..I was waiting for more comments, but I'll update as you suggested now..

ianhong:
@Praburaj and @pranavkm 

I updated per comments

src/MusicStore/Components/AnnouncementComponent.cs
```diff
@@ -10,18 +10,23 @@ namespace MusicStore.Components
...
+        [Activate]
+        public MusicStoreContext DbContext
+        {
+            get;
+            set;
+        }
 
-        public AnnouncementComponent(MusicStoreContext dbContext, IMemoryCache memoryCache)
+        [Activate]
+        public IMemoryCache MemoryCache
```
Praburaj: Could you rename the `MemoryCache` property names to `Cache`. In future we can use different cache implementations for the same sample on different environments.

src/MusicStore/Controllers/ManageController.cs
```diff
@@ -119,25 +116,25 @@ public IActionResult AddPhoneNumber()
         // POST: /Manage/DisableTwoFactorAuthentication
         [HttpPost]
         [ValidateAntiForgeryToken]
-        public async Task<IActionResult> DisableTwoFactorAuthentication()
+        public async Task<IActionResult> DisableTwoFactorAuthentication(CancellationToken cancellationToken)
```
Praburaj: Can we rename the variable name to `requestAborted` to indicate what this token is for?

Praburaj: Same comment in all applicable places.

src/MusicStore/Controllers/StoreController.cs
```diff
@@ -9,32 +9,35 @@ namespace MusicStore.Controllers
...
+        [Activate]
+        public MusicStoreContext DbContext
+        {
+            get;
+            set;
+        }
 
-        public StoreController(MusicStoreContext context, IMemoryCache memoryCache)
+        [Activate]
+        public IMemoryCache MemoryCache
```
Praburaj: Same comment here. Rename the property name to `Cache`.

src/MusicStore/Controllers/ManageController.cs
ianhong: @Praburaj 
They were originally requestAborted, but it has been changed to cancellationToken based on @pranavkm 's suggestion. 

@pranavkm  Actually, I agree with @Praburaj. I understand cancellationToken should be a common name for that type, but in this case requestAborted should more understandable to people who are not familiar to that type. Let me know your opinion on this. I can revert them back

src/MusicStore/Components/AnnouncementComponent.cs
ianhong: @Praburaj 
I think MemoryCache is not bad name because it is of IMemoryCache..I think we can change the name when we have to change its type to something else. What do you think?

Praburaj: MemoryCache is not bad :smile: But this can be general like `_cache` like it is now.

ianhong: @Praburaj Well. maybe we can use Cache just like using dbContext for MusicStoreContext?? I can change them to Cache if @pranavkm agree.

yishaigalatzer: Why moving these to activates properties makes things better?

-----Original Message-----

ianhong: For example, It has two actions. One of them doesn't need the memory cache to run. It it is a constructor parameter, users need to mock it even fo the action that doesn't require the cache.

src/MusicStore/Controllers/AccountController.cs
```diff
@@ -38,7 +35,7 @@ public IActionResult Login(string returnUrl = null)
         [HttpPost]
         [AllowAnonymous]
         [ValidateAntiForgeryToken]
-        public async Task<IActionResult> Login(LoginViewModel model, string returnUrl = null)
+        public async Task<IActionResult> Login(CancellationToken cancellationToken, LoginViewModel model, string returnUrl = null)
```
pranavkm: Move to end.

src/MusicStore/Controllers/ManageController.cs
pranavkm: Don't have a particularly strong opinion on this. That said, the param name `cancellationToken` is ubiquitous in all APIs and in theory the request being aborted is not the only reason you might want to pass in a CancellationToken to an application.

pranavkm:
:shipit: when @Praburaj's happy.

src/MusicStore/Controllers/ManageController.cs
Praburaj: But this token just indicates if request was aborted. Otherwise it will not be clear on what token is passed around here.

Praburaj:
:shipit: After these comments addressed

ianhong:
2e2ca59aa50f6491ce8c1ba8c8b96787bd9d997a

PR #396, opened 2015-02-20
React to AuthN changes

HaoK:
@Praburaj 

Verified that manually external login works for Twitter


src/MusicStore/Controllers/AccountController.cs
```diff
@@ -383,7 +383,7 @@ public ActionResult ExternalLogin(string provider, string returnUrl = null)
         [ValidateAntiForgeryToken]
         public async Task<ActionResult> ExternalLoginConfirmation(ExternalLoginConfirmationViewModel model, string returnUrl = null)
         {
-            if (User.Identity.IsAuthenticated)
+            if (User.IsLoggedIn())
```
Praburaj: Should we call this method `IsSignedIn()` just to be consistent with `SignInManager`, `SignIn` and `SignOut` methods in various places?

HaoK: Yeah probably, I'll switch to IsSignedIn

Praburaj:
>Verified that manually external login works for Twitter

Other ones should be similar. But worth validating and running the tests once.

HaoK:
I would expect they basically never ever pass on my machine :/

HaoK:
Updated with renames, (ClaimsIdentity extensions were moved to IPrincipal)

Praburaj:
:)

Praburaj:
Looks good. :shipit:

HaoK:
Updated with more renames

PR #404, opened 2015-02-26
MusicStore Unit Tests (1) - MVC, EF, Cache, RequestAborted, User

ianhong:
@Praburaj, @rynowak, and @Eilon 

I'm opening a new PR for dev

Please review this unit tests. (
 These unit tests shows examples of unit tests for MVC, EF, Cache, FormCollection, User, RequestAborted 
 Didn't do much formatting like the rest of the MusicStore codes.

I'm still working on 
 1. Session
 2. UserManager
 3. SignInManager
 4. Cookies


msotclas:
    <p>
    </p>

TTYL, MSOTBOT;

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+using Microsoft.AspNet.Http.Core.Collections;
+using Microsoft.AspNet.Mvc;
+using Microsoft.AspNet.Routing;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.DependencyInjection.Fallback;
+using Xunit;
+using MusicStore.Controllers;
+using MusicStore.Models;
+
+namespace UnitTests
```
rynowak: Please read and follow our guidelines for naming of test projects and namespaces

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+{
+    public class CheckoutControllerTest
+    {
+        IServiceProvider _serviceProvider;
```
rynowak: access modifier

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            services.AddEntityFramework()
+                      .AddInMemoryStore()
+                      .AddDbContext<MusicStoreContext>();
+
+            _serviceProvider = services.BuildServiceProvider();
+        }
+
+        [Fact]
+        public void AddressAndPaymente_ReturnsDefaultView()
```
rynowak: typo

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+            Assert.Null(viewResult.ViewName);
+        }
+
+        [Fact]
+        public void AddressAndPayment_ReturnsOrderIfInvalidPromoCode()
+        {
+            // Arrange
+            var context = new DefaultHttpContext();
+            context.Request.Form =
+                new FormCollection(new Dictionary<string, string[]>());
```
rynowak: Why is this needed?

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+                ActionContext = new ActionContext(
+                    context,
+                    new RouteData(),
+                    new ActionDescriptor()),
+            };
+
+            var order = new Order();
+
+            // Act
+            var result = controller.AddressAndPayment(order, new CancellationToken(false)).Result;
```
rynowak: Don't use `.Result`, make the test async instead

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            var controller = new CheckoutController()
+            {
+                ActionContext = new ActionContext(
+                    context,
+                    new RouteData(),
+                    new ActionDescriptor()),
+            };
+
+            var order = new Order();
```
rynowak: Should this have some actual data in it?

test/UnitTests/CheckoutControllerTest.cs
rynowak: Ah I see, the promo-code is read from the form. A comment here would be helpful.

Also this is missing a test where the promo code is used.

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            // Assert
+            var viewResult = Assert.IsType<ViewResult>(result);
+
+            Assert.NotNull(viewResult.ViewData);
+            Assert.Same(order, viewResult.ViewData.Model);
+        }
+
+        [Fact]
+        public void AddressAndPayment_ReturnsOrderIfRequestCanceled()
```
rynowak: Looking at the code in music store, this is a kinda strange use of the cancellation token. Let's leave this test as-is. We might have to discuss this more.

test/UnitTests/CheckoutControllerTest.cs
rynowak: Also notice that this action mutates the order and puts it in the DB - there's no verification of that.

rynowak: The controller code here that's getting the PromoCode from the form data directly should be model binding it using `[FromForm]`

test/UnitTests/HomeControllerTest.cs
```diff
@@ -0,0 +1,130 @@
...
+using Xunit;
+using MusicStore.Controllers;
+using MusicStore.Models;
+using System.Linq;
+
+namespace UnitTests
+{
+    public class HomeControllerTest
+    {
+        IServiceProvider _serviceProvider;
```
Praburaj: readonly.

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            // Assert
+            var viewResult = Assert.IsType<ViewResult>(result);
+
+            Assert.NotNull(viewResult.ViewData);
+            Assert.Equal(orderId, viewResult.ViewData.Model);
+        }
+
+        [Fact]
+        public void Complete_RetrunsErrorIfInvalidOrder()
```
rynowak: typo

test/UnitTests/CheckoutControllerTest.cs
Praburaj: and readonly.

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        public CheckoutControllerTest()
+        {
+            var services = new ServiceCollection();
+
+            services.AddEntityFramework()
+                      .AddInMemoryStore()
+                      .AddDbContext<MusicStoreContext>();
+
+            _serviceProvider = services.BuildServiceProvider();
```
rynowak: This is a bit of strange use of a service provider - I'm assuming this was the guidance from EF about how to test with DbContext?

What's the model here with In Memory store? How long do the changes live?

test/UnitTests/HomeControllerTest.cs
```diff
@@ -0,0 +1,130 @@
...
+            var controller = new HomeController();
+            var errorView = "~/Views/Shared/Error.cshtml";
+
+            // Act
+            var result = controller.Error();
+
+            // Assert
+            var viewResult = Assert.IsType<ViewResult>(result);
+
+            Assert.Equal(errorView, viewResult.ViewName);
```
rynowak: Looking at the comments for this action - we don't have any plans to implement `[HandleError]` that should be part of a middleware. We should remove this action.

rynowak:
Some good progress here :watch:

rynowak:
@Praburaj is there a reason we're not checking ModelState.IsValid in `AddressAndPayment`? This will let in invalid data.

test/UnitTests/CheckoutControllerTest.cs
ianhong: I understand that, but I followed the same pattern in MusicStore. Since the namespace of the existing E2ETests is just E2ETests, I used UnitTests. If you are ok, I'd like to clean up this namespace later when we decide the namespace guideline for MusicStore.

test/UnitTests/CheckoutControllerTest.cs
ianhong: Since I asserted the same object, I don't think we need data in Order..
            Assert.Same(order, viewResult.ViewData.Model);

ianhong: But, I added a comment here like 
// Do not need actual data for Order; the Order object will be checked for the reference equality.

test/UnitTests/CheckoutControllerTest.cs
ianhong: That's my understanding after discussing with @lukewaters. The model is defined in the MusicStoreContext class. I haven't thought about its lifetime though. 
@lukewaters, could you add a little bit explains here??

test/UnitTests/HomeControllerTest.cs
ianhong: Thanks. I'll make sure to remove this test when we remove the corresponding action.

test/UnitTests/CheckoutControllerTest.cs
ianhong: @rynowak 
that's correct, i missed the test, and it will be the same test to check the DB status change from order. But..that code path uses ShoppingCart, which uses Session. I didn't include that test in this PR. I'll make sure to add that test after figuring out how to handle Session.

test/UnitTests/CheckoutControllerTest.cs
lukewaters: In our tests we usually combine this into one statement, something like:
```
            ServiceProvider
                = new ServiceCollection()
                    .AddEntityFramework()
                    .AddInMemoryStore()
                    .ServiceCollection()
                    .BuildServiceProvider();
```

lukewaters: this is how we generate DbContexts for tests, but you should ensure that they are disposed by doing something like this

```
            var options = new DbContextOptions()
                .UseModel(model);
            options.UseInMemoryStore();

            using (var context = new DbContext(ServiceProvider, options))
            {
            }
```

ianhong: Thanks @lukewaters I'll update the tests accordingly. How about the lifetime of InMemoryStore?? Is it per process or something else?

lukewaters: It should be managed by the DbContext.  As long as you dispose the context appropriately you shouldn't have an issue

Praburaj:
@rynowak - yes it needs to validate modelstate. I think some of this is a copy from the old projects. Probably missed there.

Praburaj:
@ianhong - do you mind adding a change validate the model state?

test/UnitTests/CheckoutControllerTest.cs
rynowak: Please follow the guidance for any new work. If the other project needs to change, it will.

test/UnitTests/CheckoutControllerTest.cs
rynowak: >  I didn't include that test in this PR. I'll make sure to add that test after figuring out how to handle Session.

Seems fine.

BTW I think the best way forward here is not to mock session, it's to mock the shopping cart. Let's discuss after completing the current round

test/UnitTests/CheckoutControllerTest.cs
ianhong: @rynowak 
I talked to @lukewaters. I think we should leave this as it is. In our case, we should register MusicStoreDbContext first. 
1) Because there is no good way to get the model object out of the MusicStoreContext
2) All the actions actually use the MusicStoreDbContext type, but if we simply have DbContext there is no good way to cast it to the type the actions need

ianhong:
@rynowak, @Praburaj and @lukewaters 
I updated this PR per your comments.

test/UnitTests/CheckoutControllerTest.cs
Praburaj: @ianhong MusicStore tests were written long back and that time I did not know about these guidelines. I have been fixing some of these while I write new code. So feel free to fix the unit tests project to follow guidelines in newer projects.

test/UnitTests/project.json
```diff
@@ -0,0 +1,16 @@
+﻿{
+  "compilationOptions": {
+    "warningsAsErrors": "true"
+  },
+  "dependencies": {
+    "MusicStore": { "type": "build", "version": "1.0.0-*" },
```
Eilon: Any reason for this to be a build-time dependency?

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+                ActionContext = new ActionContext(
+                    context,
+                    new RouteData(),
+                    new ActionDescriptor()),
+            };
+
+            var order = new Order();
+
+            // Act
+            var result = controller.AddressAndPayment(order, new CancellationToken(true)).Result;
```
davidfowl: CanellationToken.None

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,168 @@
...
+            dbContext.Add(new Order()
+            {
+                OrderId = orderId,
+                Username = userName
+            });
+            dbContext.SaveChanges();
+
+            var controller = new CheckoutController()
+            {
+                ActionContext = new ActionContext(
```
davidfowl: This looks pretty boiler plate...

test/UnitTests/project.json
ianhong: No, I changed it to default

test/UnitTests/CheckoutControllerTest.cs
ianhong: Did you mean to use CancellationToken.None instead of Cancellationtoken(f'alse)..I changed them accordingly.

Eilon:
Not sure if this was mentioned somewhere else but the project / folder name needs to be renamed. A project named "UnitTests" is not sufficiently descriptive. Unit tests *for what*? The pattern we use elsewhere is `<main-project-name>.Test`.

ianhong:
@Eilon 
I'll update this project name once I have more feedbacks.

MusicStore.sln
```diff
@@ -24,6 +24,8 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "MusicStore", "src\MusicStor
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "MusicStore.Spa", "src\MusicStore.Spa\MusicStore.Spa.kproj", "{93891170-A8D5-46FD-A291-40F90CF258C2}"
 EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "UnitTests", "test\UnitTests\UnitTests.kproj", "{EDCE05A6-BF3B-45AC-A2C9-CFDB5CD8F57D}"
```
rynowak: This should be named differently. Please look at the guidelines

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,189 @@
...
+            var order = new Order();
+
+            // Act
+            var result = await controller.AddressAndPayment(order, new CancellationToken(false));
+
+            // Assert
+            var viewResult = Assert.IsType<ViewResult>(result);
+            Assert.NotNull(viewResult.ViewData);
+            Assert.Same(order, viewResult.ViewData.Model);
+        }
```
rynowak: Should we check that the view path is null here?

test/UnitTests/CheckoutControllerTest.cs
rynowak: Yep. You should be able to do that by doing something like 
```
var controller = new CheckoutController();
controller.HttpContext.User = ...
```
```

File: test/UnitTests/HomeControllerTest.cs [hunk 17]
```diff
@@ -0,0 +1,129 @@
...
+            var controller = new HomeController()
+            {
+                DbContext = _serviceProvider.GetRequiredService<MusicStoreContext>(),
+                Cache = _serviceProvider.GetRequiredService<IMemoryCache>(),
+            };
+
+            PopulateData(controller.DbContext);
+
+            // Action
+            var result = controller.Index().Result;
```
rynowak: Don't use `.Result` in these tests, make the test async

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                ActionContext = new ActionContext(
+                    new DefaultHttpContext(),
+                    new RouteData(),
+                    new ActionDescriptor()),
+                DbContext = dbContext,
+            };
+
+            // Act
+            var result = await controller.Complete(100);
```
rynowak: What is the `100` here? suggest named parameter

test/UnitTests/CheckoutControllerTest.cs
rynowak: Should we be disposing the services?

ianhong: Discussed with @bricelam we don't have to manually dispose this.

src/MusicStore/Controllers/CheckoutController.cs
```diff
@@ -29,8 +29,13 @@ public IActionResult AddressAndPayment()
 
         [HttpPost]
         [ValidateAntiForgeryToken]
-        public async Task<IActionResult> AddressAndPayment(Order order, CancellationToken requestAborted)
+        public async Task<IActionResult> AddressAndPayment([FromForm] Order order, CancellationToken requestAborted)
         {
+            if(!ModelState.IsValid)
```
Eilon: Formatting

test/UnitTests/CheckoutControllerTest.cs
```diff
@@ -0,0 +1,189 @@
...
+        public void AddressAndPayment_ReturnsDefaultView()
+        {
+            // Arrange
+            var controller = new CheckoutController();
+
+            // Act
+            var result = controller.AddressAndPayment();
+
+            // Assert
+            var viewResult = Assert.IsType<ViewResult>(result);
```
Eilon: Is this assert needed? Doesn't the type *have* to be a ViewResult for this to even compile? (In particular, the next assert)

Eilon: Oh sorry nevermind I get it now!

ianhong:
@Eilon , @rynowak , and @Praburaj 
I updated per comments. Please review and sign off if possible. 

Thank you

Praburaj:
I have no other comments except one. Can we just wire up logging so that the different components involved log?

ianhong:
@Eilon and @rynowak 
Just talked to @Praburaj and I don't think we need enable the logger for this type of unit tests. And, he basically signed off. Please take a look at this, when you have time.

Eilon:
:shipit:

ianhong:
5f95406fd0fb68afa11c21ef5e62dcfeb2753440

PR #406, opened 2015-02-27
Moved IConfiguration Configuration.

heathyates:
As per coding standards. Field declarations and methods with accessors should be placed at the beginning of a class. I feel this could enhance the quality of the canonical flagship example. While this point is minor, I deeply appreciate your attention and consideration. It makes it much easier to read.


msotclas:
Hi __@heathyates__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla2.msopentech.com.
    </span>

TTYL, MSOTBOT;

heathyates:
Please leave this request open for at least 4 days. I need some time to obtain the cla-required authorization. Also, I only want to add src/MusicStore/Startup.cs. Please disregard the other two changes. They were made through automation by VS15.

Praburaj:
@heathyates - could you send the PR to dev branch instead of master? Code will eventually enter all branches from here.

src/MusicStore/Startup.cs
```diff
@@ -16,6 +16,8 @@ namespace MusicStore
 {
     public class Startup
     {
+        public IConfiguration Configuration { get; private set; }
```
Eilon: The pattern we use nearly everywhere is:

1. Fields
2. Constructors
3. Properties
4. Events (which we almost never have these days)
5. Methods
6. Nested types

In my mind the rationale for this is that that's the sequence of things that execute in a class: fields are initialized, constructors are called, properties/events are set/wired-up, methods are executed, and the nested types are used as implementation details.

src/MvcMusicStore.Spa/MvcMusicStore.Spa.csproj
```diff
@@ -18,9 +18,9 @@
     <MvcBuildViews>false</MvcBuildViews>
     <UseIISExpress>true</UseIISExpress>
     <IISExpressSSLPort />
-    <IISExpressAnonymousAuthentication />
-    <IISExpressWindowsAuthentication />
-    <IISExpressUseClassicPipelineMode />
+    <IISExpressAnonymousAuthentication>disabled</IISExpressAnonymousAuthentication>
```
Eilon: Why are these changed?

heathyates: They were automatic. I have removed them from the PR I will make to Dev. As I said, I know this is a minor point and I appreciate your in depth talk. I know technically that it is an method with accessors, but for readability, I put basic methods with accessor below fields to enhance readability of a class. Regardless of the outcome of the PR, I'm a super fan of you @Eilon and @Praburaj

PR #407, opened 2015-02-27
Adding a MusicStore unit test for for SignInManager and UserManager

ianhong:
@rynowak , @Praburaj , @Eilon , and @HaoK 

Could you please review this PR??


msotclas:
    <p>
    </p>

TTYL, MSOTBOT;

ianhong:
Adding @lukewaters for EntityFramework.Identity

test/UnitTests/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        [Fact]
+        public async Task Index_ReturnsViewBagMessagesExpected()
+        {
+            // Arrange
+            var userId = "TestUserA";
+            var phone = "abcdefg";
+            var claims = new List<Claim> { new Claim(ClaimTypes.NameIdentifier, userId) };
+
+            var userManager = _serviceProvider.GetService<UserManager<ApplicationUser>>();
```
HaoK: GetRequiredService?

test/UnitTests/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public ManageControllerTest()
+        {
+            var services = new ServiceCollection();
+            services.AddEntityFramework()
+                    .AddInMemoryStore()
+                    .AddDbContext<MusicStoreContext>();
+
+            services.AddIdentity<ApplicationUser, IdentityRole>()
+                    .AddEntityFrameworkStores<MusicStoreContext>();
```
ianhong: @HaoK 
For MusicStore or any other typical MVC applications, I don't think customers need to implement a mock user store like https://github.com/aspnet/Identity/tree/dev/test/Microsoft.AspNet.Identity.InMemory.Test
because most of the time, EntityFrameworkStore should be the default user store. I think this should be good enough?? I'm wondering your opinions here

test/UnitTests/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+        [Fact]
+        public async Task Index_ReturnsViewBagMessagesExpected()
+        {
+            // Arrange
+            var userId = "TestUserA";
+            var phone = "abcdefg";
+            var claims = new List<Claim> { new Claim(ClaimTypes.NameIdentifier, userId) };
+
+            var userManager = _serviceProvider.GetService<UserManager<ApplicationUser>>();
+            await userManager.CreateAsync(
```
HaoK: Check that result succeeded

test/UnitTests/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+                new ApplicationUser
+                { Id = userId, UserName = "Test", TwoFactorEnabled = true, PhoneNumber = phone },
+                "Pass@word1");
+
+            var signInManager = _serviceProvider.GetService<SignInManager<ApplicationUser>>();
```
HaoK: GetRequiredService

test/UnitTests/ManageControllerTest.cs
HaoK: This is fine, I wouldn't bother implementing a custom store for music store, we have tests for that in the identity repo.

heathyates:
@ianhong Thanks for making these changes.

test/UnitTests/ManageControllerTest.cs
ianhong: But I still think we should provide a default constructor + some reasonable set of virtual methods for UserManager. It will make it much easier to write tests where the target actions call the following services of UserManager. 
@HaoK, what do you think?? 

                HasPassword = await UserManager.HasPasswordAsync(user),
                PhoneNumber = await UserManager.GetPhoneNumberAsync(user),
                TwoFactor = await UserManager.GetTwoFactorEnabledAsync(user),
                Logins = await UserManager.GetLoginsAsync(user),

HaoK: Talk to @divega, I tried... :)

HaoK: Basically after talking to both Diego and Lou, consensus seems to be we should be pushing people to always go through DI, so tests have to add helpers making this easy, rather than putting a separate non DI code path default logic in our framework code.  I still don't personally agree with this...but I couldn't push the boulder up the hill any longer by myself :)

test/UnitTests/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+            var viewResult = Assert.IsType<ViewResult>(result);
+            Assert.NotNull(viewResult.ViewData);
+
+            var model = Assert.IsType<IndexViewModel>(viewResult.ViewData.Model);
+            Assert.True(model.TwoFactor);
+            Assert.Equal(phone, model.PhoneNumber);
+            Assert.True(model.HasPassword);
+        }
+
+        private class TestHttpContextAcccessor : IHttpContextAccessor
```
ianhong: I think UserManager is a bit easier to handle, but SignInMager seems much more difficult to mock or control in unit tests. I had to impelment IHttpContextAccessor and HttpContext to mock the following call:
await SignInManager.IsTwoFactorClientRememberedAsync(user)

To do this, I had to look into the source codes of SignInMager, DefaultHttpContext, and so on. I think this should be too much to end customers. I think SignInManager should be more difficult to use in unit test scenarios than UserManager. 

Again, I think SignInMager should provide something like the default constructor + the reasonable set of virtual methods or callbacks..

test/UnitTests/ManageControllerTest.cs
ianhong: yes.but..I think we could make something like the default constructor of DefaultHttpContext. The default constructor has been added later to support the unit test scenarios..

test/UnitTests/project.json
```diff
@@ -0,0 +1,16 @@
+﻿{
+  "compilationOptions": {
+    "warningsAsErrors": "true"
+  },
+  "dependencies": {
+    "MusicStore": { "type": "build", "version": "1.0.0-*" },
```
Eilon: Why "build" here too?

test/UnitTests/project.json
```diff
@@ -0,0 +1,16 @@
...
+    "Microsoft.AspNet.Mvc" : "6.0.0-*",
+    "xunit.runner.kre": "1.0.0-*"
+  },
+  "commands": {
+    "test": "xunit.runner.kre"
+  },
+  "frameworks": {
+    "aspnet50": { }
```
Eilon: Does this not run on `aspnetcore50`?

ianhong: Sorry, I didn't pay attention when I copied this from an MVC unit test.. I updated this.

ianhong:
Updated per @Eilon and @HaoK 's comments

test/MusicStore.Test/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+            var phone = "abcdefg";
+            var claims = new List<Claim> { new Claim(ClaimTypes.NameIdentifier, userId) };
+
+            var userManager = _serviceProvider.GetRequiredService<UserManager<ApplicationUser>>();
+            var userManagerResult = await userManager.CreateAsync(
+                new ApplicationUser { Id = userId, UserName = "Test", TwoFactorEnabled = true, PhoneNumber = phone },
+                "Pass@word1");
+            Assert.True(userManagerResult.Succeeded);
+
+            var signInManager = _serviceProvider.GetRequiredService<SignInManager<ApplicationUser>>();
```
ianhong: @HaoK 
Based on your comments, I think UserManager should look fine. How about SignInManager? Can this be a recommended way to use in the unit test scenario? Is there any better way?

test/MusicStore.Test/project.json
```diff
@@ -0,0 +1,17 @@
+﻿{
+  "compilationOptions": {
+    "warningsAsErrors": "true"
+  },
+  "dependencies": {
+    "MusicStore": { "type": "default", "version": "1.0.0-*" },
```
Eilon: You can just remove all the special stuff from here. Just `"MusicStore": "",` I think should work (it's a project-to-project reference).

test/MusicStore.Test/ManageControllerTest.cs
HaoK: Looks fine, although you could consider Mocking instead of creating Test classes, see https://github.com/aspnet/Identity/blob/dev/test/Microsoft.AspNet.Identity.Test/SignInManagerTest.cs#L125

ianhong: @HaoK 
I understand that, but that wasn't a goal of this sample. We wanted to let users can do unit tests without using moq. I think that's why we have the default constructor for defaulthttpcontext, and some other MVC components like ViewData. I think we should discuss this more.

By the way, please sign this PR out, if possible.

test/MusicStore.Test/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
+﻿using System;
+using System.Collections.Generic;
```
rynowak: license header

ianhong: got it. The other files in this MusicStore have no license header, and I'll add the license header for the new files, but let me update the existing files in a seperate PR.

Praburaj: I vaguely remember license header is not required for samples. @Eilon is the license header required for the samples?

rynowak: Ok, well trust what @Eilon says. I tend to put it on everything #YOLO

test/MusicStore.Test/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+using Microsoft.AspNet.Http.Core;
+using Microsoft.AspNet.Http.Core.Security;
+using Microsoft.AspNet.Http.Security;
+using Microsoft.AspNet.Identity;
+using Microsoft.AspNet.Mvc;
+using Microsoft.AspNet.Routing;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.DependencyInjection.Fallback;
+using Xunit;
+using MusicStore.Models;
```
rynowak: sort :laughing:

test/MusicStore.Test/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            var httpContext = _serviceProvider.GetRequiredService<IHttpContextAccessor>().Value;
+            httpContext.User = new ClaimsPrincipal(new ClaimsIdentity(claims));
+
+            var controller = new ManageController()
+            {
+                ActionContext = new ActionContext(
+                    httpContext,
+                    new RouteData(),
+                    new ActionDescriptor()),
```
rynowak: Similar to the feedback on the other PR, this can just be

`var controller = new ManageController()
{
  UserManager...
  SignInManager...
};
controller.ActionContext.HttpContext = httpContext;

test/MusicStore.Test/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+                    .AddEntityFrameworkStores<MusicStoreContext>();
+
+            // IHttpContextAccessor is required for SignInManager, and UserManager
+            services.AddInstance<IHttpContextAccessor>(new TestHttpContextAcccessor());
+
+            _serviceProvider = services.BuildServiceProvider();
+        }
+
+        [Fact]
+        public async Task Index_ReturnsViewBagMessagesExpected()
```
rynowak: I don't see any verification here of any view-bag messages. What am I missing?

test/MusicStore.Test/ManageControllerTest.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        private class TestHttpContext : DefaultHttpContext
+        {
+            public override async Task<IEnumerable<AuthenticationResult>>
+                AuthenticateAsync(IEnumerable<string> authenticationTypes)
+            {
+                return await Task.Run<IEnumerable<AuthenticationResult>>(
+                    () =>
+                     new AuthenticateContext(authenticationTypes).Results);
```
rynowak: use `Task.FromResult` (and drop the `async`) when you need to return sync data from a Task api.

rynowak:
I got the impression before looking at this that it would be really complicated, and it's actually quite simple. The only thing I don't love about it is that a user wanting to write this code has to know about `IHttpContextAccessor` and has to set it up manually. If we can find a small tweak to remove that necessity then this will really *really* **really** be very simple in terms of cost benefit. @HaoK thoughts?

This is really more of an integration test (it uses the **real** auth services to hit the DB), but IMO that's a better fit for testing auth anyway.

HaoK:
@rynowak So SignInManager really needs an HttpContext as that's where the AuthFeature hangs, so what kind of tweaks did you have in mind?  We can certainly expose the HttpContext property setter on the manager, but unless IHttpContextAccessor becomes an optional service in the constructor that doesn't really help that much since the test is going thru the service collection to get the identity components.

rynowak:
@HaoK does adding auth to the service provider add `IHttpContextAccessor`?

HaoK:
@rynowak No IHttpContextAccessor only comes from hosting, as the hosting engine is what sets the HttpContext typically

HaoK:
We could just use the built in HttpContextAccessor from hosting, the test code would then look something like:

```
  var accessor = new HttpContextAccessor();
  accessor.Context = <appropriate test http context>;
  services.AddInstance<IHttpContextAccessor>(accessor);
  ```

**Comment by ianhong** [comment]
@rynowak and @HaoK 
The codes itself look simple, but my concern is in order to write this type of test mocking SignInManager's methods, users should know their implementation details too. For example, to mocking SignInManager.IsTwoFactorClientRememberAsync method, users should know the IsTwoxx method calls HttpContext.AuthenticateAsync. And it should have much more than that method 

I hate this, but I think I should give up this PR, and I should use Moq for at least SignInManager for MusicStore as @HaoK suggested before. 

Please let mek now your opinions

**Comment by ianhong** [comment]
@HaoK 
I just talked to @rynowak. We decided to moving forward this PR, and make it happen. I'll send another PR using Moq later for comparision..

File: test/MusicStore.Test/ManageControllerTest.cs (hunk 10 above)
Comment (Eilon) [inline test/MusicStore.Test/ManageControllerTest.cs]: For these types of samples we generally haven't bothered.

File: test/MusicStore.Test/ManageControllerTest.cs [hunk 15]
```diff
@@ -0,0 +1,116 @@
...
+            var viewResult = Assert.IsType<ViewResult>(result);
+            Assert.NotNull(viewResult.ViewData);
+
+            var model = Assert.IsType<IndexViewModel>(viewResult.ViewData.Model);
+            Assert.True(model.TwoFactor);
+            Assert.Equal(phone, model.PhoneNumber);
+            Assert.True(model.HasPassword);
+        }
+
+        private class TestHttpContextAcccessor : IHttpContextAccessor
```
Eilon: Can this whole class can just be this?
```
    public HttpContext Value { get; set; } = new TestHttpContext();
```

Eilon: Though BTW I think we also just renamed the `Value` property to be `Context` or something so keep an eye on that before merging.

Praburaj: Yeah its `HttpContext` now.

ianhong:
@HaoK, @rynowak , and @Eilon 

I updated this PR per @HaoK 's suggestion. Instead of TestHttpCotnextAccessor, I used the default implementation of it and made the test codes simpler. Please take a look at this, and sign it off if possible.

HaoK:
Looks :shipit: to me

Eilon:
:shipit: seems good for now. I'm sure at some point we'll take another high-level look anyway.

ianhong:
6efdbc75c77927247d755067bad9b7c008f40f9a

PR #410, opened 2015-03-02
Add precompiler to allow kre to precompile views

shhsu:
Verified by running kpm build and observed __ASP_ASSEMBLY.__PreGeneratedViewCollection.ctor() has registered all cshtml

### DIFF
--- a/src/MusicStore.Spa/compiler/preprocess/RazorPreCompilation.cs
+++ b/src/MusicStore.Spa/compiler/preprocess/RazorPreCompilation.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using Microsoft.AspNet.Mvc;
+using Microsoft.Framework.Runtime;
+
+namespace MusicStore
+{
+    public class RazorPreCompilation : RazorPreCompileModule
+    {
+        public RazorPreCompilation(IServiceProvider provider,
+                                   IApplicationEnvironment applicationEnvironment)
+            : base(provider)
+        {
+            GenerateSymbols = string.Equals(applicationEnvironment.Configuration,
+                                            "debug",
+                                            StringComparison.OrdinalIgnoreCase);
+        }
+    }
+}
--- a/src/MusicStore/compiler/preprocess/RazorPreCompilation.cs
+++ b/src/MusicStore/compiler/preprocess/RazorPreCompilation.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using Microsoft.AspNet.Mvc;
+using Microsoft.Framework.Runtime;
+
+namespace MusicStore
+{
+    public class RazorPreCompilation : RazorPreCompileModule
+    {
+        public RazorPreCompilation(IServiceProvider provider,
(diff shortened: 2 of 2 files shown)


msotclas:
Hi __@shhsu__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Peter Hsu). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

Praburaj:
Looks good to me. @shhsu - do you need the same for the SPA project as well?

Tragetaschen:
The path to `ICompileModule`s is still [compiler/preprocess](https://github.com/aspnet/XRE/blob/dev/src/Microsoft.Framework.Runtime.Hosting/ProjectFilesCollection.cs#L14), not with capital letters.

This won't work on Linux.

Praburaj:
@Tragetaschen - good point.

shhsu:
Is that a bug on XRE?

Praburaj:
@shhsu and this may break the build on the current version of mono due to this bug https://github.com/aspnet/XRE/issues/1048. We may have to wait on this change.

shhsu:
OK, let me know if this is unblocked

src/MusicStore.Spa/Compiler/Preprocess/RazorPreCompilation.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
Praburaj: License header is not required for this sample. Can we remove this license header?

src/MusicStore/Compiler/Preprocess/RazorPreCompilation.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
Praburaj: Same here?

shhsu:
Updated

shhsu:
@Tragetaschen

Should we change this PR to fit the dnx code or should dnx code be changed to fix this problem? We are going against naming convention here if we make folder name lower case

Praburaj:
VS Templates seems to create `Compiler\Preprocess` folders. And on windows the file under this folder is recognized as preprocess files even though the globbing is case sensitive. 

@troydai is doing some investigation on how this is working.

Praburaj:
https://github.com/aspnet/filesystem/issues/88 is filed to fix globbing to make it case sensitive. I also filed a template bug to fix VS templates to produce `compiler\preprocess` folders with lower case.

shhsu:
updated

Praburaj:
:shipit:

PR #414, opened 2015-03-04
Re-enable the Ntlm authentication tests

Praburaj:
User.IsSignedIn() does not return true for NTLM. So pulling the NTLM check out of the code block.

### DIFF
--- a/src/MusicStore/Views/Shared/_LoginPartial.cshtml
+++ b/src/MusicStore/Views/Shared/_LoginPartial.cshtml
@@ -2,30 +2,27 @@
 
 @if (User.IsSignedIn())
 {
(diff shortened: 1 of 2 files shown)


Praburaj:
@HaoK

Praburaj:
@HaoK - can we extend `User.IsSignedIn()` to return `true` for `WindowsIdentity` as well?

HaoK:
@blowdart What do you think?  I added a new ClaimsPrincipal extension for Identity, currently it only looks for identities generated by Identity, I'm not sure it makes sense to couple WindowsIdentity logic to identity, as typically you wouldn't want to pull in Identity just to use the extension method...

blowdart:
Yea, it feels weird to extend it to a WindowsIdentity

PR #423, opened 2015-03-09
Music Store sample Unit Tests - Final

ianhong:
@Praburaj, @Eilon,

This is the rest of the sample unit tests for MusicStore. Please review this. One new thing in this PR is an unit test for AntiForgery for the ShoppingCart controller.

Adding @harshgMSFT for AntiForgery


msotclas:
    <p>
    </p>

TTYL, MSOTBOT;

test/MusicStore.Test/ShoppingCartControllerTest.cs
```diff
@@ -27,6 +30,8 @@ public ShoppingCartControllerTest()
                       .AddInMemoryStore()
                       .AddDbContext<MusicStoreContext>();
 
+            services.AddMvc();
```
ianhong: Added Mvc for AntiForgery. Instead of adding entire MVC, we can also add a set of services needed as follows:

            services.AddDataProtection();
            services.AddSingleton<IClaimUidExtractor, DefaultClaimUidExtractor>();
            services.AddSingleton<AntiForgery, AntiForgery>();
            services.AddSingleton<IAntiForgeryAdditionalDataProvider, DefaultAntiForgeryAdditionalDataProvider>();

@harshgMSFT and @Praburaj,
Which one do you think we should recommend to customers?

Praburaj: `AddMvc` looks lot simpler and familiar than adding these many lines which requires to know about service types.

ianhong: Thanks @Praburaj 
I found MusicStore don't use DI for AntiForgery. 
https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Controllers/ShoppingCartController.cs#L74-L75

I'm wondering if I could change this to use the property injection like the other services, cache and dbcontext??

Praburaj:
@ianhong sure. fine with me.

test/MusicStore.Test/ShoppingCartControllerTest.cs
```diff
@@ -135,6 +140,107 @@ public ShoppingCartControllerTest()
...
+                Session = CreateTestSession(),
+            };
+
+            var httpContext = new DefaultHttpContext();
+            httpContext.SetFeature<ISessionFeature>(sessionFeature);
+            httpContext.Session.SetString("Session", "CartId_A");
+
+            // Creates the albums of AlbumId = 1 ~ 10.
+            var dbContext = _serviceProvider.GetRequiredService<MusicStoreContext>();
+            var alubms = CreateTestAlbums(itemPrice: 10);
```
Eilon: Typo

test/MusicStore.Test/ShoppingCartControllerTest.cs
Eilon: Shouldn't that use `[ValidateAntiForgeryToken]`? In https://github.com/aspnet/MusicStore/commit/ca515e3e875b97e9327aefc1646fdca018e12ae2 it looks like some actions used the attribute and some used some custom code... how odd...

ianhong: I think it is because this particular action uses a custom java script to post data. 
https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Views/ShoppingCart/Index.cshtml#L27

@Praburaj, could you clairify this?

Eilon: Oh weird... why would it need special codez? Doesn't the browser automatically include the XSRF token?

harshgMSFT: I think it is because they want to test the ajax scenario, (which is not a mainline usecase, but there are some people who wanted to do it.). For ajax scenario a token has to be explicitly passed.

Eilon: I was so sure that the browser will automatically include the cookies... are you sure that isn't the case?

harshgMSFT: cookies yes, but not the form token. For the antiforgery validation to work you need both the cookie and the form token to be present and match.

harshgMSFT: And in this particular case i think that he does not use the oob ValidateAntiForgeryToken because it needs to do custom parsing.

Eilon: Oh, yes, the form token. But then if it's passed in as part of the form body (is it?), won't the regular validation attribute accept it? Or is it getting sent in a query string?

ianhong: I think the regular  validation attribute shoudln't work. The values are sent in a custom format with javascript.
https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Views/ShoppingCart/Index.cshtml#L9-L13

Praburaj: @Eilon this view does not have a form. The click even of the remove link is hijacked to perform a POST. So the ajax code sends the xsrf token along with the request. I remember annotating the action with [ValidateAntiForgery..] did not work for some reason and I saw this way of validation in one of the web api samples in asp.net site. But I think I can replace it with the attribute as long as I POST the token in the expected format. I'll give it a try.

Praburaj: @harshgMSFT - the javascript is posting the xsrf in this format right now (cookietoken:formtoken). Will posting the same format work with the attribute or is it required in a different format?

test/MusicStore.Test/CheckoutControllerTest.cs
```diff
@@ -43,6 +48,66 @@ public void AddressAndPayment_ReturnsDefaultView()
...
         [Fact]
+        public async Task AddressAndPayment_RedirectToCompleteWhenSuccessful()
+        {
+            // Arrange
+            var httpContext = new DefaultHttpContext();
+
+            var orderId = 10;
+            var order = new Order()
+            {
+                OrderId = orderId,
```
Praburaj: nit: can remove the comma at the end.

Praburaj:
Rest looks good to me.

test/MusicStore.Test/CheckoutControllerTest.cs
ianhong: In another review, I was told that leaving , is better since we don't have to touch the line in case we have to add another line. I think it could be a personal preference. I'd like to leave , here for now. :-)

test/MusicStore.Test/ShoppingCartControllerTest.cs
harshgMSFT: Nope that would not work. The OOB attribute only understands a form value of ```__RequestVerificationToken```. 
What we should do : 
Use the ```[ValidateAntiForgeryToken]``` and in our ajax request we should post with a key of ```__RequestVerificationToken``` and a value of just the form token. That should work and should be secure.

harshgMSFT: you can also prevent vulnerability by adding this as a header instead of a form value.

ianhong:
@Praburaj and @Eilon 
I updated this PR to react to 29909e15f33b22db1fe638457315199e0a46eeea. Let me know if there is anything I should do more. 

Thanks

test/MusicStore.Test/StoreControllerTest.cs
```diff
@@ -0,0 +1,175 @@
...
+            Assert.Equal(genreName, viewModel.Name);
+            Assert.NotNull(viewModel.Albums);
+            Assert.Equal(3, viewModel.Albums.Count);
+        }
+
+        [Fact]
+        public async Task Details_ReturnsHttpNotFoundWhenNoAlbumData()
+        {
+            // Arrange
+            var albumId = 0;
```
Praburaj: Do we know for sure that `0` will not be generated as id? Instead is it safe to use some negative values?

test/MusicStore.Test/StoreControllerTest.cs
```diff
@@ -0,0 +1,175 @@
...
+            Assert.NotNull(genre);
+            Assert.NotNull(genre.Albums.SingleOrDefault(a => a.AlbumId == albumId));
+            Assert.Null(viewModel.Artist);
+
+            var cachedAlbum = cache.Get<Task<Album>>("album_1");
+            Assert.NotNull(cachedAlbum);
+            Assert.Equal(albumId, cachedAlbum.Result.AlbumId);
+        }
+
+        private Genre[] CreateTestGenres(int numberOfGenres, int numberOfAlbums, DbContext dbContext)
```
Praburaj: The return value of `CreateTestGenres` seems not used any where. Can this be changed to `private void`?

Praburaj: oh.. I see one usage where the value is used. Ignore this comment.

Praburaj: `private static` ?

Praburaj:
Looks good to me. :shipit:

ianhong:
1b502cbee7b86709a0b9a0afd4de4f3c97db52ae

PR #430, opened 2015-03-16
Ef : Enabled Autogeneration for ArtistId & GenreId using sequence

smitpatel:
Recent change in EF requires the principal root in chain of foreignkey to have GenerateValueOnAdd enabled.
Due to this GenreId & ArtistId requires GenerateValueOnAdd as true. This is done by conventions automatically.
Since explicit values are being inserted in both the columns using sequence instead of Identity (which is also default by convention)


smitpatel:
@Praburaj @AndriySvyryd

AndriySvyryd:
:shipit:

PR #438, opened 2015-03-24
Fixing a build break in .net 4.5 version of projects

suhasj:
@Praburaj 

We have this one unmerged commit from release to dev. Need to verify if this is good to merge


Praburaj:
Dev branch already has this change : https://github.com/aspnet/MusicStore/commit/a7a70786a5084f3d1da52fffb61532ad6425f0c5. So we should not require this change.

PR #439, opened 2015-03-25
Remove k command and use dnx instead

ChengTian:
https://github.com/aspnet/dnx/issues/1473
@davidfowl @Praburaj @muratg

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,17 +14,17 @@ This project is part of ASP.NET 5.0. You can find samples, documentation and get
 ## Run on WebListener/Kestrel:
 * Open a command prompt and cd `\src\MusicStore\`.
 * **[WebListener]:**
-	4. Run `k web` (Application started at URL **http://localhost:5002/**).
+	4. Run `dnx . web` (Application started at URL **http://localhost:5002/**).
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -14,17 +14,17 @@ This project is part of ASP.NET 5.0. You can find samples, documentation and get
 ## Run on WebListener/Kestrel:
 * Open a command prompt and cd `\src\MusicStore\`.
 * **[WebListener]:**
-	4. Run `k web` (Application started at URL **http://localhost:5002/**).
+	4. Run `dnx . web` (Application started at URL **http://localhost:5002/**).
```
Praburaj: why does it have a `.`?

ChengTian: It's appbase. We are not simply renaming `k.cmd` to `dnx.cmd`, we are discarding `k.cmd` and use `dnx.exe`, which is invoked by `k.cmd`. `dnx.exe` has different usage.

hishamco: @Praburaj we should use dnx . web and dnx .test to run and test the current application

Praburaj: sounds good. Thanks!

PR #447, opened 2015-03-29
React to hosting

HaoK:
https://github.com/aspnet/Hosting/pull/210/files

Does music store actually use the ApplicationName for anything?  There's no easy way to set app name now.

cc @davidfowl @Praburaj

### DIFF
--- a/src/MusicStore/Program.cs
+++ b/src/MusicStore/Program.cs
@@ -17,14 +17,9 @@ namespace MusicStore
             var config = new Configuration();
             config.AddCommandLine(args);
 
-            var context = new HostingContext()
-            {
-                Configuration = config,
-                ServerFactoryLocation = "Microsoft.AspNet.Server.WebListener",
-                ApplicationName = "MusicStore"
-            };
-
-            using (new HostingEngine().Start(context))
+            using (WebApplication.CreateHostingEngine(config)
+                .UseServer("Microsoft.AspNet.Server.WebListener")
+                .Start())
             {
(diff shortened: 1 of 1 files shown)


Praburaj:
>Does music store actually use the ApplicationName for anything? There's no easy way to set app name now.

Not really.

src/MusicStore/Program.cs
```diff
@@ -17,14 +17,9 @@ public Task<int> Main(string[] args)
...
 
-            var context = new HostingContext()
-            {
-                Configuration = config,
-                ServerFactoryLocation = "Microsoft.AspNet.Server.WebListener",
-                ApplicationName = "MusicStore"
-            };
-
-            using (new HostingEngine().Start(context))
+            using (WebApplication.CreateHostingEngine(config)
```
Praburaj: Can't we bring back `WebApp.Start<T>`?

HaoK: T is the startup type?  @lodejard suggested that belongs as a TestServer extension instead, I can bring it up with him again

Praburaj: >T is the startup type

yeah.

HaoK: I'm not sure what change you are advocating for music store though,  you don't like how hosting automatically scans for startup today?  You would prefer something more explicit like WebApplication.CreateHostingEngine(config).Start<Startup>()?

Praburaj: oh no. That question was in general. Not specific to MusicStore.

PR #454, opened 2015-04-01
Fixing the logging package names in the unit test projects.

Praburaj:


Praburaj:
@victorhurdugaci @Eilon @ianhong

davidfowl:
:shipit:

victorhurdugaci:
Looks good. But we should make the CI fail first by cleaning those network shares that have old packages.

Praburaj:
>But we should make the CI fail first by cleaning those network shares that have old packages.

Yeah.

PR #463, opened 2015-04-15
Bundle and deploy

victorhurdugaci:
Few changes for testing on Nano server:
- Enable the MusicStore tests to run on CoreCLR
- Script that bundles the application, copies it to the remote machine, sets up a remote PS session and then runs the tests

Please review @Praburaj , @pranavkm

### DIFF
--- a/test/E2ETests/project.json
+++ b/test/E2ETests/project.json
@@ -21,6 +21,14 @@
                 "System.Net.Http": "",
                 "System.Xml": ""
             }
+        },
+        "dnxcore50": {
+            "dependencies": {
+                "Microsoft.Win32.Primitives": "4.0.0-beta-*",
+                "System.Data.SqlClient": "4.0.0-beta-*",
+                "System.Net.Http": "4.0.0-beta-*",
+                "System.Xml.XmlDocument": "4.0.0-beta-*"
+            }
         }
     }
 }
--- a/tools/BundleAndDeploy.ps1
+++ b/tools/BundleAndDeploy.ps1
@@ -1,0 +1,90 @@
+param(
+    [Parameter(Mandatory=$true)]
+    [string] $projectFile,
+    [Parameter(Mandatory=$true)]
+    [string] $server,
+    [string] $serverFolder,
+    [string] $userName,
+    [string] $password
+)
+
+function UpdateHostInProjectJson($projectFile, $newhost)
+{
+    (Get-Content $projectFile) | 
+    Foreach-Object {
+        $_ -replace "http://localhost", "http://$newhost"
+    } |
+    Set-Content $projectFile 
+}
+ 
+if (-not (Test-Path $projectFile)) {
+    Write-Error "Couldn't find $projectFile"
+    exit 1
+}
+
+$projectName = (get-item $projectFile).Directory.Name
+$workDir = (get-item $projectFile).Directory.FullName
+$remoteRoot="\\" + $server
+$remoteDir= Join-Path $remoteRoot -ChildPath $serverFolder
+
+Write-Host "Project: $projectName"
+Write-Host "Work directory: $workDir"
+Write-Host "Remote dir: $remoteDir"
+Write-Host "User name: $userName"
+
+try
+{
+    if ($userName) {
+		net use $remoteRoot $password /USER:$userName
+        if ($lastexitcode -ne 0) {
+            exit 1
+        }
+    }
+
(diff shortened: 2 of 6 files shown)


victorhurdugaci:
@moozzyk you might be interested in this if you try to test something

test/E2ETests/project.json
```diff
@@ -21,6 +21,14 @@
                 "System.Net.Http": "",
                 "System.Xml": ""
             }
+        },
+        "dnxcore50": {
```
Praburaj: Is there a way to disable running the coreclr version of the library on teamcity? I think the build.cmd runs once against all the targets configured.

Praburaj: running it on coreclr is a bit redundant on teamcity and its going to take more time.

test/E2ETests/project.json
```diff
@@ -21,6 +21,14 @@
...
+            "dependencies": {
+                "Microsoft.Win32.Primitives": "4.0.0-beta-*",
```
Praburaj: Win32 Primitives might not be required anymore.

tools/BundleAndDeploy.ps1
```diff
@@ -0,0 +1,90 @@
...
+    [string] $serverFolder,
+    [string] $userName,
+    [string] $password
+)
+
+function UpdateHostInProjectJson($projectFile, $newhost)
+{
+    (Get-Content $projectFile) | 
+    Foreach-Object {
+        $_ -replace "http://localhost", "http://$newhost"
```
Praburaj: How do you set this appbase url in the test?

moozzyk:
@victorhurdugaci looks like I was able to successfully bundle and deploy to nano server with this script. 

running MusicStore on nano failed but I don't think it is because of this script:
```
[10.135.33.3]: PS C:\dev\MusicStore> .\run
.\run : System.DllNotFoundException: Unable to load DLL 'sni.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
    + CategoryInfo          : NotSpecified: (System.DllNotFo...LT: 0x8007007E):String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError

   at System.Data.SqlClient.SNINativeMethodWrapper.UnmanagedIsTokenRestricted(IntPtr token, Boolean& isRestricted)

 at System.Data.Win32NativeMethods.IsTokenRestrictedWrapper(IntPtr token)
   at System.Data.ProviderBase.DbConnectionPoolIdentity.GetCurrent()
```

victorhurdugaci:
Oh yes, the SQL client doesn't work. You have to use the in memory store

tools/BundleAndDeploy.ps1
```diff
@@ -0,0 +1,90 @@
...
+
+Write-Host "Project: $projectName"
+Write-Host "Work directory: $workDir"
+Write-Host "Remote dir: $remoteDir"
+Write-Host "User name: $userName"
+
+try
+{
+    if ($userName) {
+		net use $remoteRoot $password /USER:$userName
```
moozzyk: nit: spaces

Praburaj:
And how do you choose the app to use inmemory store on nano server?

victorhurdugaci:
For now you edit [this line](https://github.com/aspnet/MusicStore/blob/master/src/MusicStore/Startup.cs#L33) :blush:

Praburaj:
Hmm :smile: Probably you could have a config override like this. We already do this for IIS scenarios. https://github.com/aspnet/MusicStore/blob/dev/test/E2ETests/compiler/shared/Mocks/StartupSocialTesting.cs#L35

victorhurdugaci:
It should use the new `IRuntimeEnvironment` to figure out if needs the in memory store

Praburaj:
oh really. Does that give the host information as well?

victorhurdugaci:
Yes, it gives:
- OS name
- OS version
- host version
- host type (clr/mono/coreclr)
- host architecture

Praburaj:
ah.. I mean does that say if it is IIS / WebListener / Kestrel as well?

victorhurdugaci:
oh, sorry. no, it only gives information about OS and DNX

test/E2ETests/project.json
victorhurdugaci: Would be okay to run these tests only on CoreCLR? They spawn a server anyway, right?

PR #470, opened 2015-04-21
onboard to Application Insights SDK for the Asp.Net v5 applications

SergKanz:
@Olegsych can you take a look?


msotclas:
Hi __@SergKanz__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Sergey Kanzhelev). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

olegsych:
:+1:

src/MusicStore/config.json
```diff
@@ -1,4 +1,7 @@
 ﻿{
+    "ApplicationInsights": {
+        "InstrumentationKey": "797d1481-9988-48f2-b41b-f4e305cd9207"
+    },
     "DefaultAdminUsername": "Administrator@test.com",
     "DefaultAdminPassword": "YouShouldChangeThisPassword1!",
```
danroth27: We should not encourage people to put passwords in config.json. We should discuss how to handle the user secrets correctly.

src/MusicStore/config.json
```diff
@@ -1,4 +1,7 @@
 ﻿{
+    "ApplicationInsights": {
+        "InstrumentationKey": "797d1481-9988-48f2-b41b-f4e305cd9207"
+    },
     "DefaultAdminUsername": "Administrator@test.com",
```
danroth27: How would this work if you don't have a username/password? We need folks to be able to clone Music Store and be able to run it with no changes.

src/MusicStore/config.json
olegsych: Good point, although we don't consider the Application Insights instrumentation keys to be secrets because they are embedded in plain text in the JavaScript instrumentation code and in the telemetry that can be sent by HTML clients over HTTP. What would be a proper way to store the admin user name and password values in this example? Can we do the same for the instrumentation key?

src/MusicStore/config.json
olegsych: Are you referring to an Azure subscription to view the Application Insights? I don't think we can enable public access to a particular Application Insights resource.

What do you think about leaving the InstrumentationKey blank with a comment explaining how to add it? This will silently disable sending of telemetry to Application Insights and allow each user to use their own subscription to configure and see telemetry.

Eilon:
I'm not sure we want to add this to Music Store at all. I think it's fine to fork Music Store and add App Insights to that fork. The idea of this sample is to compare & contrast to the previous MVC version of Music Store. There should generally be no differences in functionality between the two.

olegsych:
@SergKanz, how do we fork aspnet/MusicStore to microsoft/ApplicationInsights-MusicStore?

SergKanz:
@Eilon we believe that having it by default in musicstore will be a good showcase of telemetry and monitoring integration with ASP.NET v5 applications. Some sort of best practices - when you include logging, also include telemetry. We are also working on adding Application Insights instructions on [docs.asp.net](http://docs.asp.net/). Having code integrated can make documenting easier.

I see your point though. There are lots of best practices out there to include them all =). However if you'll change your mind - let us know.

@olegsych I think we have enough documentation so we don't need to maintain a fork of musicstore. It is very easy to onboard it for demos/testing anyway.

olegsych:
OK. I'll delete the olegsych/MusicStore fork then to avoid maintaining dead code.

Eilon:
@SergKanz we also use this app in perf tests so keeping the ASP.NET 5 version as close as possible to ASP.NET 4.x is important (we don't use this exact app, but rather a fork of it, but nevertheless having to tweak it all the time is error prone).

PR #474, opened 2015-04-24
Adding the unit tests for ViewComponents in MusicStore with the default ViewContext

ianhong:
@Praburaj, @Eilon, and @rynowak 

I added more unit tests with the default ViewContext constructor with aspnet/Mvc#2442


msotclas:
    <p>
    </p>

TTYL, MSOTBOT;

test/MusicStore.Test/AnnouncementComponentTest.cs
```diff
@@ -0,0 +1,107 @@
...
+            _serviceProvider = services.BuildServiceProvider();
+        }
+
+        [Fact]
+        public async Task AnnouncementComponent_Returns_LatestAlbum()
+        {
+            // Arrange
+            var viewContext = new ViewContext()
+            {
+                ViewData = new ViewDataDictionary(
```
ianhong: This ViewData initialization seems very common here. We can add this type of code to the ViewContext.ViewData property like https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Controller.cs#L162-L187

What do you think??

test/MusicStore.Test/AnnouncementComponentTest.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
```
Eilon: Add file headers.

test/MusicStore.Test/AnnouncementComponentTest.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            // Assert
+            Assert.NotNull(result);
+            var viewResult = Assert.IsType<ViewViewComponentResult>(result);
+            Assert.Null(viewResult.ViewName);
+            var albumResult = Assert.IsType<Album>(viewResult.ViewData.Model);
+            Assert.Equal(today, albumResult.Created.Date);
+        }
+
+        private void PopulateData(DbContext context, DateTime latestAlbumDate)
```
Eilon: Static?

test/MusicStore.Test/AnnouncementComponentTest.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            var announcementComponent = new AnnouncementComponent()
+            {
+                DbContext = _serviceProvider.GetRequiredService<MusicStoreContext>(),
+                Cache = _serviceProvider.GetRequiredService<IMemoryCache>(),
+                ViewContext = viewContext,
+                ViewData = new ViewDataDictionary(viewContext.ViewData),
+            };
+
+            var today = DateTime.UtcNow.Date;
```
Eilon: Recommend using an arbitrary constant date here to ensure there is never variability in this test.

test/MusicStore.Test/CartSummaryComponentTest.cs
```diff
@@ -0,0 +1,109 @@
...
+            var sessionFeature = new SessionFeature()
+            {
+                Session = CreateTestSession(),
+            };
+            viewContext.HttpContext.SetFeature<ISessionFeature>(sessionFeature);
+            viewContext.HttpContext.Session.SetString("Session", cartId);
+
+            // DbContext initialization
+            var dbContext = _serviceProvider.GetRequiredService<MusicStoreContext>();
+            PopulateData(dbContext, cartId, albumTitle: "AblumA", itemCount: 10);
```
Eilon: Typo in the album name :smile:

test/MusicStore.Test/CartSummaryComponentTest.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                ViewData = new ViewDataDictionary(
+                    new EmptyModelMetadataProvider(),
+                    new ModelStateDictionary()),
+                HttpContext = new DefaultHttpContext()
+            };
+
+            // Session initialization
+            var cartId = "CartId_A";
+            var sessionFeature = new SessionFeature()
```
Eilon: I looked as Music Store app itself, and to my surprise, it doesn't seem to be using session state at all. It seems to be doing a "hand-rolled" session state by creating its own cookie and using that to store the cart id.

Does this unit test work just as well without all this session stuff? My guess is yes. However, that would also be surprising because that means that something else is missing here...

test/MusicStore.Test/GenreMenuComponentTest.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            // Assert
+            Assert.NotNull(result);
+            var viewResult = Assert.IsType<ViewViewComponentResult>(result);
+            Assert.Null(viewResult.ViewName);
+            var genreResult = Assert.IsType<List<Genre>>(viewResult.ViewData.Model);
+            Assert.Equal(9, genreResult.Count);
+        }
+
+        private void PopulateData(MusicStoreContext context)
```
Eilon: Static?

test/MusicStore.Test/CartSummaryComponentTest.cs
ianhong: @Eilon 
Did you take a look at this?? I think MusicStore using the session feature. 
https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Models/ShoppingCart.cs#L150-L164

test/MusicStore.Test/AnnouncementComponentTest.cs
ianhong: Agree, but I think I should use this here due to how the component is implemented
https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Components/AnnouncementComponent.cs#L39-L46
Let me know if you want me to change this component so that it simply returns the latest one.

test/MusicStore.Test/AnnouncementComponentTest.cs
ianhong: MusicStore files don't have file headers, and I remember that you were fine with it. But, let me know if you want me to add file headers for all MusicStore files..

Eilon: Oh you're right, ignore this.

test/MusicStore.Test/CartSummaryComponentTest.cs
Eilon: Oh, I see, you're right. I got confused from there being two different things called "Session" in there, plus the comments in that code are incorrect (it mentions cookies, which is irrelevant).

test/MusicStore.Test/AnnouncementComponentTest.cs
Eilon: Hmm, this might be a case where the product code has to change to be unit testable. Code that uses DateTime.Now (or similar) cannot be unit tested. You need to use something like https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/ISystemClock.cs to create a fake known time.

For example, I think this test could fail unexpectedly if the test run started at 11:59:59.9999pm and the date rolled over because it became 12:00:00.0000am the next day when the rest of the code executed, then the calculations would be off by 1 entire day.

ianhong: got it. I'll update the MusicStore code to use that interface too..

src/MusicStore/Components/AnnouncementComponent.cs
```diff
@@ -25,6 +26,13 @@ public IMemoryCache Cache
             set;
         }
 
+        [Activate]
+        public ISystemClock SystemClock
```
ianhong: @Eilon 
There is two identical system clock from Microsoft.AspNet.Authentication and  Microsoft.Framework.Caching.Memory.Infrastructure. 
@Praburaj suggested me to use one that under Caching. I think we should combine them at some point.

ianhong:
@Eilon 
I updated this PR for review. 

Thanks

src/MusicStore/Components/AnnouncementComponent.cs
Eilon: We should redefine the interface here. The one from caching is effectively an internal implementation detail.

test/MusicStore.Test/AnnouncementComponentTest.cs
```diff
@@ -0,0 +1,117 @@
...
+
+                return albums;
+            }
+        }
+
+        private class TestSystemClock : ISystemClock
+        {
+            private DateTime _date;
+
+            public TestSystemClock(DateTime date)
```
Eilon: Why take in a DateTime only to then convert to a DateTimeOffset?

test/MusicStore.Test/AnnouncementComponentTest.cs
```diff
@@ -0,0 +1,117 @@
...
+                        Created = latestAlbumDate.AddDays(1 - n),
+                    }).ToArray();
+
+                return albums;
+            }
+        }
+
+        private class TestSystemClock : ISystemClock
+        {
+            private DateTime _date;
```
Eilon: `readonly`

test/MusicStore.Test/AnnouncementComponentTest.cs
```diff
@@ -0,0 +1,117 @@
...
+                      .AddDbContext<MusicStoreContext>();
+
+            _serviceProvider = services.BuildServiceProvider();
+        }
+
+        [Fact]
+        public async Task AnnouncementComponent_Returns_LatestAlbum()
+        {
+            // Arrange
+            var today = DateTime.Parse("10/30/2012");
```
Eilon: You can just construct a DateTime (or DateTimeOffset, really) instead of parsing. Parsing is affected by culture, which we need to avoid.

src/MusicStore/Components/AnnouncementComponent.cs
ianhong: @Eilon I'll use it from the authentication one then..

test/MusicStore.Test/AnnouncementComponentTest.cs
ianhong: This is because the type of Album's Create property is DateTime, but ISystemClock interace returns DateTimeOffset
https://github.com/aspnet/MusicStore/blob/master/src/MusicStore/Models/Album.cs#L36-L38
So, they shoud be converted at some point either here or when populating test data. 
Generally we should use DateTimeOffset over DateTime, but I think it is fine with the Album model since we are concerned about Date for this model. (https://msdn.microsoft.com/en-us/library/bb384267(v=vs.110).aspx)

Let me know if you want to convert it when populating test data instead of in this interface.

src/MusicStore/Components/AnnouncementComponent.cs
Eilon: No, just redefine it. Don't use it from any other assembly because these interfaces are not contractual.

test/MusicStore.Test/AnnouncementComponentTest.cs
Eilon: This is why we need to redefine the interface for the specific scenarios of this app.

ianhong:
Hi @Eilon 
I updated per comments..

Thanks

src/MusicStore/Components/SystemClock.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Provides access to the normal system clock.
+    /// </summary>
+    public class SystemClock : ISystemClock
+    {
+        /// <inheritdoc />
+        public DateTime UtcNow
+        {
+            get
+            {
+                // The clock measures whole seconds only, to have integral expires_in results, and
```
Eilon: Is this milliseconds stuff needed here? I think the Authentication version of this had that because the time gets persisted into various other places such as cookies or something.

ianhong: This is not necessary here, but miliisecond time resolution should be different depening on SQL server versions, products, or sometimes on its configruation. And Album data should be stored in DB. So, I thought it's ok to keep this code here. But let me know if you want to remove this part, and simply return dattime..

src/MusicStore/Components/ISystemClock.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace MusicStore.Components
+{
+    /// <summary>
+    /// Abstracts the system clock to facilitate testing.
+    /// </summary>
+    public interface ISystemClock
```
ianhong: @Eilon 
I added this interface, and the method UtcNow returns DateTime, but...I'm thinking this name is not good here or we should return DateTimeOffset from this interface. This interface should be used not only for Album, but it also could be used for the other models where DateTimeOffset is more apporiate. 
I'd like to change its method to return DateTimeOffset instead of DateTime..What do you think??

src/MusicStore/Components/SystemClock.cs
Eilon: Eh ok perhaps it's fine to leave it (it's not wrong, it just looks weird).

However, please update the comment to avoid mention of "expires_in" because that's an auth concept that is irrelevant here.

src/MusicStore/Components/ISystemClock.cs
Eilon: Not sure I understand what's wrong with the name... can you clarify that?

As far as switching between DateTime and DateTimeOffset, if we do such a change, we'd want to do it in the entire project. That might be better suited as a separate change and for now have this clock type use whatever date/time concept is already used in the app.

src/MusicStore/Components/SystemClock.cs
ianhong: Got it.. I'll update the comment as follows:
// The clock measures whole seconds only, and truncates the milliseconds,
// because the millisecond resolutions are inconsistent among various underlying systems.
Since this is only thing I should upate from this PR so far, could you sign it off, if there is no more comments?  I'll update the comment when I check in this PR.

Eilon: Looks fine but in the second sentence "the resolutions are" should be "resolution is" (no "the").

Eilon:
:shipit: after the last comment fix-up.

davidfowl:
@rynowak Whats the deal with this?

kichalla:
797f8029064fc218b0c0d6ba8cc85d937ca697c3

PR #478, opened 2015-05-06
Use the in memory store on the Nano server

victorhurdugaci:
The SQL client doesn't work well on the Nano server so for now we'll use the in memory store.

Please review @suhasj

### DIFF
--- a/src/MusicStore/Startup.cs
+++ b/src/MusicStore/Startup.cs
@@ -17,7 +17,7 @@ namespace MusicStore
 {
     public class Startup
     {
-        private readonly IRuntimeEnvironment _runtimeEnvironment;
+        private readonly Platform _platform;
 
         public Startup(IApplicationEnvironment env, IRuntimeEnvironment runtimeEnvironment)
         {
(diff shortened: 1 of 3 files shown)


src/MusicStore/Startup.cs
```diff
@@ -36,8 +36,9 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.Configure<AppSettings>(Configuration.GetSubKey("AppSettings"));
 
-            //Sql client not available on mono
-            var useInMemoryStore = _runtimeEnvironment.RuntimeType.Equals("Mono", StringComparison.OrdinalIgnoreCase);
+            // Use the in memory store when running on mono or on the Nano server
+            var useInMemoryStore = _runtimeEnvironment.RuntimeType.Equals("Mono", StringComparison.OrdinalIgnoreCase) ||
```
suhasj: Can you move the check condition for Mono in the Platform class ? You can replace this usage through out the project making it consistent

victorhurdugaci:
@suhasj updated

suhasj:
Can you make the changes in this file too https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/StartupOpenIdConnect.cs

suhasj:
:shipit:

Issue #488, opened 2015-05-18
enable Kestrel tests


suhasj:
Kestrel based tests are failing currently. Need to enable them once fixed

kichalla:
I re-enabled the tests.

Issue #487, opened 2015-05-18
Enable publish test


suhasj:
Publish tests are failing currently due to a regression in DNX. Disabling the tests as of now and tracking to enable them when fixed.

PR #489, opened 2015-05-18
Disabling tests to unblock CI

suhasj:
Disabling tests

### DIFF
--- a/test/E2ETests/PublishAndRunTests.cs
+++ b/test/E2ETests/PublishAndRunTests.cs
@@ -15,7 +15,8 @@ namespace E2ETests
     {
         [ConditionalTheory, Trait("E2Etests", "E2Etests")]
         [FrameworkSkipCondition(RuntimeFrameworks.Mono)]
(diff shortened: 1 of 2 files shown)


test/E2ETests/PublishAndRunTests.cs
```diff
@@ -15,7 +15,8 @@ public class PublishAndRunTests_OnX64
     {
         [ConditionalTheory, Trait("E2Etests", "E2Etests")]
         [FrameworkSkipCondition(RuntimeFrameworks.Mono)]
-        [InlineData(ServerType.WebListener, RuntimeFlavor.Clr, RuntimeArchitecture.x64, "http://localhost:5025/", false)]
+	// https://github.com/aspnet/MusicStore/issues/487
```
Eilon: Tab...

Eilon:
Looks like some of these could be disabled with `[Skip]` or something no? And for the ones that can't, can you log a bug to re-enable them?

suhasj:
Sure I have opened two bugs that are tracking the disabled tests and assigned to folks. I'll updated the spacing to be consistent

#488 
#487

Eilon:
Ah ok thanks!

PR #492, opened 2015-05-21
React to aspnet/Mvc#2489 fix

dougbu:
- rename _GlobalImport.cshtml file to _ViewImports.cshtml


Eilon:
:ski: :relaxed: - :hocho: (figure it out)

PR #498, opened 2015-06-11
Using [NotNull] and 'nameof' operator

hishamco:


Eilon:
User code doesn't use `[NotNull]` - it's currently a framework-only feature.

hishamco:
What about ```nameof``` while it's language feature?

Eilon:
Yeah `nameof()` is fine.

PR #501, opened 2015-06-18
Add AzureAD

BrennanConroy:
@kichalla @Eilon 
For release branch


Eilon:
Assuming this is needed, it's fine. I wonder if the AzureAD feed should go above NuGet though.

PR #533, opened 2015-09-03
Fix tests

BrennanConroy:
@victorhurdugaci


victorhurdugaci:
I guess that's a good change. What happened to that comma? :shipit:

PR #534, opened 2015-09-11
React to ConfigureCors api removal

kichalla:
@HaoK

### DIFF
--- a/src/MusicStore/Startup.cs
+++ b/src/MusicStore/Startup.cs
@@ -13,6 +13,7 @@ using Microsoft.Framework.Logging;
 using Microsoft.Dnx.Runtime;
 using MusicStore.Components;
 using MusicStore.Models;
+using Microsoft.AspNet.Cors.Core;
 
 namespace MusicStore
 {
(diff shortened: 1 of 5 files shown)


HaoK:
MVC adds cors I take it?  Maybe its just best to call the AddCors method instead?

kichalla:
@HaoK right, MVC adds Cors too...regarding calling AddCors instead I am not sure...actually this triggered some thought...this could apply to any service in general...
Let's assume I call AddCors(with some option settings) and then AddMVC or any other service which also does AddCors but different options settings. Now since AddCors always does a TryAdd, the services are not overriden but in case of options the final options instance can overwrite the earlier ones...so probably that not desired?

HaoK:
So its a bit muddled now, but Add now does two things, it will ALWAYS configure the options, and TryAdd the services.  So all of the options tweaking will always run.  

For this situation though I think its fine to just call AddCors explicitly since it reads better?

HaoK:
In regards to the sideffects, really nothing has changed, Configure has the same behavior as calling Add does, both of them take effect in the order you call them.

kichalla:
@HaoK thanks, updated it now

kichalla:
bfe1de4819377f25ae0452280b8b06e9f7bfaacf

PR #540, opened 2015-09-23
added missing reference to System.Data.Entity which causes build to fail.

redwards510:
The line UseGlobalApplicationHostFile just added itself after I built, so I assume that's ok.

### DIFF
--- a/src/MvcMusicStore/MvcMusicStore.csproj
+++ b/src/MvcMusicStore/MvcMusicStore.csproj
@@ -21,6 +21,7 @@
     <IISExpressWindowsAuthentication />
     <IISExpressUseClassicPipelineMode />
     <WebGreaseLibPath>..\..\packages\WebGrease.1.5.2\lib</WebGreaseLibPath>
+    <UseGlobalApplicationHostFile />
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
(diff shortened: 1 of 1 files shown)


Eilon:
Does this happen in the `dev` branch? That's the most current branch. The `master` branch here is for `beta7` so I want to make sure there isn't a mismatch here.

redwards510:
I haven't tried the dev branch. I just cloned the repository so I could try playing around with this project and it fails to build because of the missing reference.

Eilon:
@kichalla canyou take a look to see if `master` is broken?

kichalla:
@Eilon based on the PR change, its a change in `MvcMusicStore.csproj`, so our builds would not pick it up?

ajcvickers:
@Eilon Are you sure this is something we want to do? System.Data.Entity is the namespace from legacy versions of EF. Unless something has changed I don't think we want MusicStore to reference this.

Eilon:
@ajcvickers it's the MvcMusicStore project (the original one using .NET 4.x) that was using EF, so I think in principal that's fine. However, we just deleted MvcMusicStore from the `dev` branch so I'm not sure it's worth fixing something that we just deleted. All we kept in `dev` was the ASP.NET 5 / EF 7 version of MusicStore.

Nevertheless, I'm curious how we didn't notice that it was broken...

But given all that, I suspect it's fine to close this PR.

ajcvickers:
@Eilon Okay, that makes sense.

Eilon:
Closing this because in 1 week this won't matter (we deleted the affected project).

PR #545, opened 2015-10-02
Remove unneeded namespaces, use new startup APIs

Eilon:
Just cleaning up the sample even further by using some APIs that were introduced a while back.

@kichalla @dougbu


dougbu:
:shipit: nice smallerizations
AppVeyor agrees

Eilon:
Thanks, @dougbu.

@rynowak FYI.

PR #548, opened 2015-10-15
Remove source reference

victorhurdugaci:
Those files are not needed and the break test publish in https://github.com/aspnet/dnx/issues/2986

@kichalla

### DIFF
--- a/src/MusicStore/project.json
+++ b/src/MusicStore/project.json
@@ -5,9 +5,6 @@
     "description": "Music store application on ASP.NET 5",
     "version": "1.0.0-*",
     "compilationOptions": { "warningsAsErrors": true, "define": [ "DEMO", "TESTING" ] },
-    "compile": [
-        "../../test/E2ETests/compiler/shared/**/*.cs"
-    ],
     "publishExclude": "*.cmd",
     "webroot": "wwwroot",
     "dependencies": {


src/MusicStore/project.json
```diff
@@ -5,9 +5,6 @@
     "description": "Music store application on ASP.NET 5",
     "version": "1.0.0-*",
     "compilationOptions": { "warningsAsErrors": true, "define": [ "DEMO", "TESTING" ] },
-    "compile": [
-        "../../test/E2ETests/compiler/shared/**/*.cs"
-    ],
```
kichalla: I am not sure the effects of removing these(like tests other than the ones testing 'publish' itself)...my idea was that if it only effecting publish tests, then we could disable them specifically...did you try running build.cmd and see if this has any effect?

victorhurdugaci: There are no tests doing publish. The issue is that I'm not able to publish the tests on Nano server because of this

kichalla: I was referring to https://github.com/aspnet/MusicStore/blob/dev/test/E2ETests/PublishAndRunTests.cs

victorhurdugaci:
This doesn't work. It breaks tests. Closing the PR and opened this to track the issue: https://github.com/aspnet/MusicStore/issues/549

PR #550, opened 2015-10-16
Move compiler shared to its own folder

victorhurdugaci:
- Moved shared files to their own folder outside of src and test


victorhurdugaci:
cc @kichalla

kichalla:
:shipit:

PR #554, opened 2015-10-21
:green_heart: React to API change in UseDatabaseErrorPage

rowanmiller:
The overload being used in the Startup.cs has swapped to a builder pattern (see https://github.com/aspnet/Diagnostics/issues/184). But rather than updating to the new signature, just swapping to the parameterless overload since it enables everything by default (same as UseDeveloperExceptionPage()).


ajcvickers:
:shipit:

rowanmiller:
Merged https://github.com/aspnet/MusicStore/commit/42a5a2821059886deeba0ade5c6e726a620d5825

PR #555, opened 2015-10-21
:green_heart: Fix more calls to UseDatabaseErrorPage

rowanmiller:
Some more calls that I missed in my previous commit. The overload being used in the templates has swapped to a builder pattern (see aspnet/Diagnostics#184). But rather than updating to the new signature, just swapping to the parameterless overload since it enables everything by default (same as UseDeveloperExceptionPage()).


ajcvickers:
:shipit:

PR #562, opened 2015-11-11
Rybrande/empty cart

ryanbrandenburg:
Needs Antiforgery #21. Make the item count of the cart change when items are removed.


Eilon:
:shipit: (once the other one is checked in)

PR #585, opened 2016-01-11
 * Use suggested Antiforgery AJAX patern.

### LINKED ISSUE #580: Update use of Antiforgery + Ajax to built-in header support
Our use of antiforgery for ajax in this sample is a bit ad-hoc. https://github.com/aspnet/MusicStore/blob/5a3c4957d9707a9901891a6b909863d1a34cc37f/src/MusicStore/Controllers/ShoppingCartController.cs#L72

We have built in support for using a header for antiforgery in rc2, we should update the sample to use our recommended pattern.

ryanbrandenburg:
Fixes #580

### DIFF
--- a/src/MusicStore/Startup.cs
+++ b/src/MusicStore/Startup.cs
@@ -1,5 +1,4 @@
 using Microsoft.AspNet.Builder;
-using Microsoft.AspNet.Hosting;
 using Microsoft.AspNet.Identity.EntityFramework;
 using Microsoft.Data.Entity;
 using Microsoft.Extensions.Configuration;
--- a/src/MusicStore/Views/ShoppingCart/Index.cshtml
+++ b/src/MusicStore/Views/ShoppingCart/Index.cshtml
@@ -6,10 +6,9 @@
 
 @functions
 {
-    public string GetAntiXsrfToken()
+    public string GetAntiXsrfRequestToken()
     {
-        var tokens = Xsrf.GetTokens(Context);
-        return tokens.CookieToken + ":" + tokens.RequestToken;
+        return Xsrf.GetAndStoreTokens(Context).RequestToken;
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/MusicStore/Views/ShoppingCart/Index.cshtml
```diff
@@ -6,10 +6,10 @@
 
 @functions
 {
-    public string GetAntiXsrfToken()
+    public string GetAntiXsrfRequestToken()
     {
         var tokens = Xsrf.GetTokens(Context);
```
rynowak: What does this do?

src/MusicStore/Startup.cs
```diff
@@ -144,11 +146,19 @@ public void ConfigureProduction(IApplicationBuilder app, ILoggerFactory loggerFa
...
-        public void Configure(IApplicationBuilder app)
+        public void Configure(IApplicationBuilder app, IAntiforgery antiforgery)
         {
+            app.Use(next => context =>
+            {
+                var requestToken = antiforgery.GetAndStoreTokens(context).RequestToken;
+                context.Response.Cookies.Append("XSRF-TOKEN", requestToken, new CookieOptions() { HttpOnly = false });
+
+                return next(context);
+            });
```
rynowak: Why is this setting the token for every request?

ryanbrandenburg:
cc @rynowak

src/MusicStore/Views/ShoppingCart/Index.cshtml
ryanbrandenburg: This is getting the XSRF token for use in AJAX below. Are you saying there's a better way to get it on the page or that we can inline it down there?

ryanbrandenburg:
:up: :date: with what we talked about @rynowak. Also I guess technically I should cc @blowdart just in case I'm going something that's not best practice.

src/MusicStore/Startup.cs
```diff
@@ -144,10 +145,10 @@ public void ConfigureProduction(IApplicationBuilder app, ILoggerFactory loggerFa
 
             app.UseExceptionHandler("/Home/Error");
 
-            Configure(app);
+            Configure(app, antiforgery);
         }
 
-        public void Configure(IApplicationBuilder app)
```
rynowak: Revert changes here

rynowak:
:shipit:

PR #599, opened 2016-03-03
Cleanup some EF TODOs

### LINKED ISSUE #597: Cleanup EF todos (where possible)
Music store has some workarounds for limitations that used to exist in EF. Cleanup these usages.

- https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Models/ShoppingCart.cs#L107
- https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Models/MusicStoreContext.cs#L21
- https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Models/MusicStoreContext.cs#L25

natemcmaster:
Fix #597

### DIFF
--- a/src/MusicStore/Components/GenreMenuComponent.cs
+++ b/src/MusicStore/Components/GenreMenuComponent.cs
@@ -26,16 +26,13 @@ namespace MusicStore.Components
 
         private async Task<List<Genre>> GetGenres()
         {
-            // TODO [EF] We don't query related data as yet, so the OrderByDescending isn't doing anything
-            //var genres = _dbContext.Genres
-            //.OrderByDescending(
-            //    g => g.Albums.Sum(
-            //    a => a.OrderDetails.Sum(
-            //    od => od.Quantity)))
-            //.Take(9)
-            //.ToList();
-
-            return await DbContext.Genres.Take(9).ToListAsync();
+            return await DbContext.Genres
+                .Include(g => g.Albums).ThenInclude(a => a.OrderDetails)
+                // TODO use nested sum https://github.com/aspnet/EntityFramework/issues/3792
+                //.OrderByDescending(
+                //    g => g.Albums.Sum(a => a.OrderDetails.Sum(od => od.Quantity)))
+                .Take(9)
+                .ToListAsync();
         }
     }
 }
--- a/src/MusicStore/Controllers/HomeController.cs
+++ b/src/MusicStore/Controllers/HomeController.cs
@@ -60,9 +60,8 @@ namespace MusicStore.Controllers
             // Group the order details by album and return
             // the albums with the highest count
 
-            // TODO [EF] We don't query related data as yet, so the OrderByDescending isn't doing anything
             return await dbContext.Albums
-                .OrderByDescending(a => a.OrderDetails.Count())
+                .OrderByDescending(a => a.OrderDetails.Count)
                 .Take(count)
                 .ToListAsync();
         }
--- a/src/MusicStore/Models/ShoppingCart.cs
+++ b/src/MusicStore/Models/ShoppingCart.cs
@@ -10,25 +10,25 @@ namespace MusicStore.Models
     public class ShoppingCart
     {
         private readonly MusicStoreContext _dbContext;
-        private string ShoppingCartId { get; set; }
+        private readonly string _shoppingCartId;
 
-        public ShoppingCart(MusicStoreContext dbContext)
+        private ShoppingCart(MusicStoreContext dbContext, string id)
         {
             _dbContext = dbContext;
+            _shoppingCartId = id;
         }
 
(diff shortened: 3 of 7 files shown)


natemcmaster:
cc @divega

divega:
@kichalla do you still own this?

src/MusicStore/Components/GenreMenuComponent.cs
```diff
@@ -26,16 +26,13 @@ public GenreMenuComponent(MusicStoreContext dbContext)
...
-            //.Take(9)
-            //.ToList();
-
-            return await DbContext.Genres.Take(9).ToListAsync();
+            return await DbContext.Genres
+                .Include(g => g.Albums).ThenInclude(a => a.OrderDetails)
+                // TODO use nested sum https://github.com/aspnet/EntityFramework/issues/3792
+                //.OrderByDescending(
+                //    g => g.Albums.Sum(a => a.OrderDetails.Sum(od => od.Quantity)))
+                .Take(9)
```
divega: Anyone knows why this is limited to 9?

natemcmaster: I believe this was arbitrary. 9 just looks good [in the menu](https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Views/Shared/Components/GenreMenu/Default.cshtml)

src/MusicStore/Models/ShoppingCart.cs
```diff
@@ -78,14 +78,14 @@ public int RemoveFromCart(int id)
 
         public void EmptyCart()
```
divega: Nit: Any reason some methods like this one that performs queries isn't async? It seems most of what I saw above is async.

src/MusicStore/Models/ShoppingCart.cs
```diff
@@ -93,9 +93,11 @@ public void EmptyCart()
         public async Task<int> GetCount()
```
divega: Nit: Remove async keyword and await.

src/MusicStore/Models/ShoppingCart.cs
```diff
@@ -93,9 +93,11 @@ public void EmptyCart()
...
-                          where cartItem.CartId == ShoppingCartId
-                          select cartItem.Count).SumAsync();
+            return await _dbContext
+                .CartItems
+                .Where(c => c.CartId == _shoppingCartId)
+                .Select(c => c.Count)
+                .SumAsync();
         }
 
         public async Task<decimal> GetTotal()
```
divega: Same. In general methods for which the implementation returns a Task at the tail don't need to be async, just return a Task. And this is showcasing a good practice :smile:

test/MusicStore.Test/Models/ShoppingCartTest.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
```
divega: Thanks for adding tests :smile:

divega:
LGTM, with just a few minor comments. @kichalla do you need to sign off on this?

src/MusicStore/Components/GenreMenuComponent.cs
kichalla: Yeah, as far as I remember and tried, its for making the UI good. Curious, does the `Take` part need to appear after `Genres` so as to limit the entries early?

src/MusicStore/Models/ShoppingCart.cs
kichalla: :+1:

src/MusicStore/Controllers/HomeController.cs
```diff
@@ -60,9 +60,8 @@ public IActionResult AccessDenied()
             // Group the order details by album and return
             // the albums with the highest count
 
-            // TODO [EF] We don't query related data as yet, so the OrderByDescending isn't doing anything
             return await dbContext.Albums
-                .OrderByDescending(a => a.OrderDetails.Count())
+                .OrderByDescending(a => a.OrderDetails.Count)
```
kichalla: does it need to 'include' orderdetails?

kichalla:
:shipit:

src/MusicStore/Components/GenreMenuComponent.cs
natemcmaster: It shouldn't affect the query's results. There might be a slight gain on InMemory but in SQL Server this will get folded into the final sql generated.

src/MusicStore/Controllers/HomeController.cs
natemcmaster: We only need to include them back if we want to materialize the Albums.OrderDetails collection. Since we don't need them in Views/Home/Index.cshtml (and because of https://github.com/aspnet/EntityFramework/issues/4695), we are ok leaving them out for now.

PR #603, opened 2016-03-08
[Reliability] Added cache timeout control

martinpf:
Added ability to control memory cache timeout or turn off the use of the cache completely.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,3 +1,33 @@
+$ErrorActionPreference = "Stop"
+
+function DownloadWithRetry([string] $url, [string] $downloadLocation, [int] $retries) 
+{
+    while($true)
+    {
+        try
+        {
+            Invoke-WebRequest $url -OutFile $downloadLocation
+            break
+        }
+        catch
+        {
+            $exceptionMessage = $_.Exception.Message
+            Write-Host "Failed to download '$url': $exceptionMessage"
+            if ($retries -gt 0) {
+                $retries--
+                Write-Host "Waiting 10 seconds before retrying. Retries left: $retries"
+                Start-Sleep -Seconds 10
+
+            }
+            else 
+            {
+                $exception = $_.Exception
+                throw $exception
+            }
+        }
+    }
+}
+
 cd $PSScriptRoot
 
 $repoFolder = $PSScriptRoot
@@ -20,7 +50,8 @@ if (!(Test-Path $buildFolder)) {
 
     $localZipFile="$tempFolder\korebuild.zip"
     
-    Invoke-WebRequest $koreBuildZip -OutFile $localZipFile
+    DownloadWithRetry -url $koreBuildZip -downloadLocation $localZipFile -retries 6
+
     Add-Type -AssemblyName System.IO.Compression.FileSystem
     [System.IO.Compression.ZipFile]::ExtractToDirectory($localZipFile, $tempFolder)
     
--- a/src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
+++ b/src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
@@ -10,6 +10,7 @@ using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Caching.Memory;
 using MusicStore.Models;
 using MusicStore.ViewModels;
+using Microsoft.Extensions.Options;
 
 namespace MusicStore.Areas.Admin.Controllers
 {
@@ -17,12 +18,14 @@ namespace MusicStore.Areas.Admin.Controllers
     [Authorize("ManageStore")]
     public class StoreManagerController : Controller
     {
-        public StoreManagerController(MusicStoreContext dbContext)
+        public StoreManagerController(MusicStoreContext dbContext, IOptions<AppSettings> appSettings)
         {
             DbContext = dbContext;
+            AppSettings = appSettings;
         }
 
         public MusicStoreContext DbContext { get; }
+        private IOptions<AppSettings> AppSettings;
 
         //
         // GET: /StoreManager/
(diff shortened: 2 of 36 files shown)


src/MusicStore/config.json
```diff
@@ -1,14 +1,15 @@
...
     },
     "DefaultAdminUsername": "Administrator@test.com",
     "DefaultAdminPassword": "YouShouldChangeThisPassword1!",
-    "Data": {
-        "DefaultConnection": {
-            // Use a shared (and running) LocalDB database when executing in IIS e.g.
-            // "Server=(localdb)\\.\\IIS_DB;Database=MusicStore;Trusted_Connection=False;MultipleActiveResultSets=true;User ID=iis_login;Password=********"
-            "Connectionstring": "Server=(localdb)\\MSSQLLocalDB;Database=MusicStore;Trusted_Connection=True;MultipleActiveResultSets=true;Connect Timeout=30;"
-        }
+  "Data": {
```
SajayAntony: Why does this have to change?

src/MusicStore/Controllers/StoreController.cs
```diff
@@ -50,22 +54,34 @@ public StoreController(MusicStoreContext dbContext)
         {
             var cacheKey = string.Format("album_{0}", id);
             Album album;
-            if (!cache.TryGetValue(cacheKey, out album))
+
+            if (AppSettings.Value.CacheTimeout > 0)
+            {
+                if (!cache.TryGetValue(cacheKey, out album))
+                {
+                    album = await DbContext.Albums
```
SajayAntony: This is code duplication. Why not just check if 
`if(cacheTimeout > 0 && cache.TryGetValue(...) `

src/MusicStore/wwwroot/web.config
```diff
@@ -9,4 +9,10 @@
     <!-- This setting is not recommended for production -->
     <add key="ASPNET_DETAILED_ERRORS" value="true" />
   </appSettings>
+  <system.webServer>
+    <handlers>
+      <add name="httpPlatformHandler" path="*" verb="*" modules="httpPlatformHandler" resourceType="Unspecified" />
+    </handlers>
+    <httpPlatform processPath="%DNX_PATH%" arguments="%DNX_ARGS%" forwardWindowsAuthToken="false" startupTimeLimit="3600" />
```
SajayAntony: Could we make this change into music store itself?

src/MusicStore/Program.cs
```diff
@@ -7,6 +7,7 @@ public static class Program
         public static void Main(string[] args)
         {
             var host = new WebHostBuilder()
+                .UseUrls("http://+:5000")
```
SajayAntony: This should be set through environment variables and not in the app
I would recommend we stick with command line `server.urls=http://*:5000` not change the app to expose publicly

src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
```diff
@@ -45,27 +48,43 @@ public StoreManagerController(MusicStoreContext dbContext)
...
-                        album,
-                        new MemoryCacheEntryOptions().SetSlidingExpiration(TimeSpan.FromMinutes(10)));
-                }
             }
 
+
             if (album == null)
             {
-                cache.Remove(cacheKey);
+                if (AppSettings.Value.CacheTimeout > 0)
```
SajayAntony: I think you need this in multiple places. 
Recommend creating something like this. 

``` csharp
class MemoryCacheExtensions
{
 public void Remove<T>(this MemoryCache cache,T cachekey,  IOptions<AppSettings> options)
 {
    if(options.Value.CacheTimeout > 0)
        cache.Remove(cacheKey);
  }
}
```

and invoke ` cache.Remove(cacheKey, options);`

DavidObando: Indeed. It looks like the expression `AppSettings.Value.CacheTimeout > 0` actually means `should we use the cache?` but it's not explicitly mentioned, in comments, or in a utility method. This is nitpicking, but consider making this more clear.

src/MusicStore/config.json
```diff
@@ -8,7 +9,8 @@
         "DefaultConnection": {
             // Use a shared (and running) LocalDB database when executing in IIS e.g.
```
DavidObando: Consider moving comments outside of the config.json file. This is a bit of a mix of concerns: config.json should only provide the configuration required, not the logic or annotations for the logic. Additionally, if in the future we change the JSON parser it might choke on comments, as they aren't part of [the spec](http://www.ietf.org/rfc/rfc7159.txt).

src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
```diff
@@ -17,12 +18,14 @@ namespace MusicStore.Areas.Admin.Controllers
...
     {
-        public StoreManagerController(MusicStoreContext dbContext)
+        public StoreManagerController(MusicStoreContext dbContext, IOptions<AppSettings> appSettings)
         {
             DbContext = dbContext;
+            AppSettings = appSettings;
         }
 
         public MusicStoreContext DbContext { get; }
+        private IOptions<AppSettings> AppSettings;
```
Eilon: This field should be of type `AppSettings` (not `IOptions<T>`), and the field name needs to be `_appSettings` (or make it a read-only property).

test/E2ETests/Implementation/Validator.cs
```diff
@@ -29,6 +37,73 @@ public partial class Validator
             _deploymentResult = deploymentResult;
         }
 
+        private Task<HttpResponseMessage> DoGetAsync(string uri)
```
SajayAntony: Do something is quite generic. Please rename to HttpGetAsync and HttpPostAsync.

build.ps1
```diff
@@ -1,3 +1,33 @@
...
+            }
+            else 
+            {
+                $exception = $_.Exception
+                throw $exception
+            }
+        }
+    }
+}
+
```
SajayAntony: @muratg shouldn't KoreBuild do this or should we modify each build script to retry?

SajayAntony:
Please rebase your branch against reliability or dev which is at a4ed79c

build.ps1
Eilon: This is already done in KoreBuild and is already in MusicStore; I'm guessing this PR just needs to be rebased or something.

muratg: Yup, it's here: https://github.com/aspnet/KoreBuild/commit/ac2b7124a44e911cbdf4d03961640fbb8142e43a

PR #605, opened 2016-03-08
 Add license headers to all files

natemcmaster:
Looks like a lot, but it's just code cleanup. Ran R# on whole project.

cc @kichalla


src/MusicStore/Components/ISystemClock.cs
```diff
@@ -1,16 +1,19 @@
...
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// See License.txt in the project root for license information
+
+using System;
 
 namespace MusicStore.Components
 {
     /// <summary>
-    /// Abstracts the system clock to facilitate testing.
+    ///     Abstracts the system clock to facilitate testing.
```
dougbu: How is this an improvement?

src/MusicStore/Controllers/AccountController.cs
```diff
@@ -151,8 +153,10 @@ public IActionResult Register()
                     // For more information on how to enable account confirmation and password reset please visit http://go.microsoft.com/fwlink/?LinkID=320771
                     // Send an email with this link
                     string code = await UserManager.GenerateEmailConfirmationTokenAsync(user);
-                    var callbackUrl = Url.Action("ConfirmEmail", "Account", new { userId = user.Id, code = code }, protocol: HttpContext.Request.Scheme);
-                    await MessageServices.SendEmailAsync(model.Email, "Confirm your account",
+                    var callbackUrl = Url.Action(
+                        "ConfirmEmail", "Account", new { userId = user.Id, code = code }, protocol: HttpContext.Request.Scheme);
+                    await MessageServices.SendEmailAsync(
+                        model.Email, "Confirm your account",
                         "Please confirm your account by clicking this link: <a href=\"" + callbackUrl + "\">link</a>");
```
dougbu: Style is generally "if you wrap, give every argument its own line." [Line 159](https://github.com/aspnet/MusicStore/pull/605/files#diff-d9ee8fb38dcc504d9616e8e6611fae7cR159) looks particularly odd.

src/MusicStore/Controllers/AccountController.cs
```diff
@@ -174,7 +178,8 @@ public IActionResult Register()
         [AllowAnonymous]
         public async Task<ActionResult> ConfirmEmail(string userId, string code)
         {
-            if (userId == null || code == null)
+            if (userId == null
+                || code == null)
```
dougbu: Why?

src/MusicStore/Startup.cs
```diff
@@ -36,41 +39,34 @@ public void ConfigureServices(IServiceCollection services)
             services.Configure<AppSettings>(Configuration.GetSection("AppSettings"));
 
             var useInMemoryStore = !_platform.IsRunningOnWindows
-                || _platform.IsRunningOnMono
-                || _platform.IsRunningOnNanoServer;
+                                   || _platform.IsRunningOnMono
+                                   || _platform.IsRunningOnNanoServer;
```
dougbu: Definitely not. Just wrap to where VS defaults.

dougbu:
- Do we need license headers in these samples?
- Rest looks like churn that I only got about a quarter of the way through. Some changes are negative.

natemcmaster:
These are the same code formatting guidelines we use on other repos via R#.

dougbu:
Most repos don't use R# and the style there is what VS defaults to.

PR #608, opened 2016-03-12
Added cache timeout control via config

martinpf:
This should add cache control and incorporate the latest feedback


src/MusicStore/config.json
```diff
@@ -1,7 +1,8 @@
 ﻿{
-    "AppSettings": {
-        "SiteTitle": "ASP.NET MVC Music Store"
-    },
+  "AppSettings": {
+    "SiteTitle": "ASP.NET MVC Music Store",
+    "CacheTimeout": "0"
```
SajayAntony: For the first commit add the original timeout. 
That way you might be able to submit the PR to dev.

src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
```diff
@@ -17,9 +19,12 @@ namespace MusicStore.Areas.Admin.Controllers
     [Authorize("ManageStore")]
     public class StoreManagerController : Controller
     {
-        public StoreManagerController(MusicStoreContext dbContext)
+        private IOptions<AppSettings> AppSettings;
```
Eilon: Engineering guidelines: name fields `_likeThis`

PR #617, opened 2016-03-23
Enable IISExpress tests

JunTaoLuo:
I am able to run these tests locally with the old packages. Will need to test with newer packages once we have a Coherence build.

cc @Tratcher @shirhatti


BrennanConroy:
Fix https://github.com/aspnet/MusicStore/blob/6c2f6d245b5148f0df926981b374684534e77338/test/E2ETests/Implementation/Validator.cs#L203 while you're doing this

Tratcher:
:shipit:

PR #620, opened 2016-03-29
Created a configurable memory cache

kichalla:
@SajayAntony @martinpf @Tratcher

### DIFF
--- a/src/MusicStore/Controllers/HomeController.cs
+++ b/src/MusicStore/Controllers/HomeController.cs
@@ -5,12 +5,19 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Options;
 using MusicStore.Models;
 
 namespace MusicStore.Controllers
 {
     public class HomeController : Controller
     {
+        private readonly AppSettings _appSettings;
+
+        public HomeController(IOptions<AppSettings> options)
+        {
+            _appSettings = options.Value;
+        }
         //
         // GET: /Home/
         public async Task<IActionResult> Index(
@@ -26,14 +33,17 @@ namespace MusicStore.Controllers
 
                 if (albums != null && albums.Count > 0)
                 {
-                    // Refresh it every 10 minutes.
-                    // Let this be the last item to be removed by cache if cache GC kicks in.
-                    cache.Set(
-                        cacheKey,
-                        albums,
-                        new MemoryCacheEntryOptions()
+                    if (_appSettings.CacheDbResults)
+                    {
+                        // Refresh it every 10 minutes.
+                        // Let this be the last item to be removed by cache if cache GC kicks in.
+                        cache.Set(
+                            cacheKey,
+                            albums,
+                            new MemoryCacheEntryOptions()
                             .SetAbsoluteExpiration(TimeSpan.FromMinutes(10))
                             .SetPriority(CacheItemPriority.High));
+                    }
                 }
             }
 
--- a/src/MusicStore/Properties/AppSettings.cs
+++ b/src/MusicStore/Properties/AppSettings.cs
@@ -3,5 +3,7 @@
     public class AppSettings
     {
         public string SiteTitle { get; set; }
+
+        public bool CacheDbResults { get; set; } = true;
     }
 }
--- a/src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
+++ b/src/MusicStore/Areas/Admin/Controllers/StoreManagerController.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Options;
 using MusicStore.Models;
 using MusicStore.ViewModels;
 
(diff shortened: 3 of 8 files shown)


NuGet.config
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetcidev/api/v3/index.json" />
+    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvolatiledev/api/v3/index.json" />
```
kichalla: ignore this change

SajayAntony:
Ignoring the nuget.config fix 
:shipit:

Tratcher:
Why??

kichalla:
@Tratcher For Perf scenarios

Tratcher:
:-1: Just implement the interface and add it to DI, there's no need to change all the callers.

kichalla:
@Tratcher I initially thought of doing so but then thought that it would affect any of our framework components which depend on `IMemoryCache` and which I believe is not what the Perf team wants.

@SajayAntony, can you confirm? When the cache needs to be disabled, its only for the user code that needs to be disabled and not the framework code itself, right?

For example, the framework depends on `IMemoryCache` service in the following places and even they would be affected. 
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs#L42
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs#L103

SajayAntony:
@kichalla yes we only care about disabling memory cache for usercode since if that is being used in the framework then we want to make sure any issues are uncovered from framework misbehaving.

Tratcher:
This still seems like a very strange abstraction. AppSettings.StoreInCache doesn't impact IMemoryCache.Get or Remove, just Set. Get rid of ConfigurableMemoryCache and inject AppSettings into StoreManagerController and check StoreInCache on line 56. Much less invasive.

src/MusicStore/Properties/AppSettings.cs
```diff
@@ -3,5 +3,7 @@
     public class AppSettings
     {
         public string SiteTitle { get; set; }
+
+        public bool StoreInCache { get; set; }
```
Tratcher: Can this default to true and be disabled via config?

Tratcher: Naming. Store what in cache? `CacheDbLookups`?

kichalla:
> This still seems like a very strange abstraction. AppSettings.StoreInCache doesn't impact IMemoryCache.Get or Remove, just Set. Get rid of ConfigurableMemoryCache and inject AppSettings into StoreManagerController and check StoreInCache on line 56. Much less invasive.

But this seems more invasive. This kind of check needs to be made in all places where the cache is being used. Currently we hide this behind the `ConfigurableMemoryCache` layer.

src/MusicStore/Properties/AppSettings.cs
kichalla: OK. Probably `UseCache` instead?

Tratcher:
It's actually less invasive as you only need to do it where you call Set, three places, as opposed to the 12 places you had to change IMemoryCache to your custom type.

src/MusicStore/Properties/AppSettings.cs
```diff
@@ -3,5 +3,7 @@
     public class AppSettings
     {
         public string SiteTitle { get; set; }
+
+        public bool UseCache { get; set; }
```
Tratcher: :-1: Meh, it still doesn't say what you're using cache for.

kichalla:
> It's actually less invasive as you only need to do it where you call Set, three places, as opposed to the 12 places you had to change IMemoryCache to your custom type.

Not sure I like it. Also considering what @SajayAntony mentioned earlier about isolating cache usage to catch issues any framework related issues, then we need to avoid using caching in all (get, set, remove) the cases when it is disabled.

Tratcher:
If you never call Set for a particular key you shouldn't ever run into issues with Get or Remove for that key. @SajayAntony?

kichalla:
ping @SajayAntony

SajayAntony:
@Tratcher @kichalla as long as we don't change framework (MVC/EFs) caching behavior by changing the DI container I'm fine either way. The only change I think we are looking to do is disable data level caching so that we execute the entire code path on every request.

Tratcher:
OK. @kichalla it should be enough to optionally avoid calling Set. You don't need a special cache implementation.

kichalla:
updated @SajayAntony @Tratcher

src/MusicStore/Controllers/HomeController.cs
```diff
@@ -26,14 +33,17 @@ public class HomeController : Controller
...
                 {
-                    // Refresh it every 10 minutes.
-                    // Let this be the last item to be removed by cache if cache GC kicks in.
-                    cache.Set(
+                    if (_appSettings.StoreInCache)
+                    {
+                        // Refresh it every 10 minutes.
+                        // Let this be the last item to be removed by cache if cache GC kicks in.
+                        cache.Set(
                         cacheKey,
```
Tratcher: indentation?

src/MusicStore/Properties/AppSettings.cs
```diff
@@ -3,5 +3,7 @@
     public class AppSettings
     {
         public string SiteTitle { get; set; }
+
+        public bool StoreInCache { get; set; } = true;
```
Tratcher: Name still needs work. Store what in cache?

Tratcher:
:shipit:

src/MusicStore/Properties/AppSettings.cs
kichalla: suggestions?

Tratcher: CacheDbResults?

PR #634, opened 2016-04-13
Minor improvements to EF usage.

anpete:
1) Queries are untracked by default.
2) Use projection query in CartSummaryComponent.
3) Removed redundant async methods.


kichalla:
:shipit:

PR #637, opened 2016-04-14
Use UseServer() instead of UseKestrel() to allow override in tests

JunTaoLuo:
This is needed to fix the build. Explanation is in the comments.

Review: @mikeharder @Tratcher 
Approval: @Eilon @DamianEdwards


Tratcher:
:shipit:

PR #638, opened 2016-04-14
Various EF usage improvements

anpete:
- Use more efficient projection query for GenreMenuComponent.
- Change some sync EF calls to async.
- Log at Information level when Development.

### DIFF
--- a/src/MusicStore/Components/GenreMenuComponent.cs
+++ b/src/MusicStore/Components/GenreMenuComponent.cs
@@ -19,20 +19,9 @@ namespace MusicStore.Components
 
         public async Task<IViewComponentResult> InvokeAsync()
         {
-            var genres = await GetGenres();
+            var genres = await DbContext.Genres.Select(g => g.Name).Take(9).ToListAsync();
 
             return View(genres);
-        }
-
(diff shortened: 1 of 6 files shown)


anpete:
ping @rynowak

src/MusicStore/Components/GenreMenuComponent.cs
```diff
@@ -19,20 +19,9 @@ public GenreMenuComponent(MusicStoreContext dbContext)
 
         public async Task<IViewComponentResult> InvokeAsync()
         {
-            var genres = await GetGenres();
+            var genres = await DbContext.Genres.Select(g => g.Name).Take(9).ToListAsync();
```
rynowak: Just leaving a thought here for posterity (not actually suggesting a change).

If in the future we do want to order these by quantity (see TODO) then we could just wrap this in an #CacheTagHelper to prevent it from reexecuting on every page load.

rynowak:
:shipit:

src/MusicStore/Components/GenreMenuComponent.cs
anpete: Will add the TODO. We should be able to do this without Includes.

PR #646, opened 2016-04-21
View full albums by genre

ralmsdeveloper:
Show albums related to gender


Eilon:
@ralmsdeveloper thanks for the PR; however at this time we are trying to keep the functionality of the Music Store relatively stable because we use it in scenarios such as perf tests and various functional tests.

ralmsdeveloper:
I thought it was interesting to put this functionality to studies, to serve as an example!

I'm looking for work!

Eilon:
@ralmsdeveloper one of the best ways to find items to work on is to find items labeled `up-for-grabs`, which the team puts on issues that should be relatively easy to approach for someone who hasn't contributed before.

Here's a query: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+label%3Aup-for-grabs+user%3Aaspnet

Keep in mind that many of these issues are currently in the Backlog, so they aren't planned for the current release. But if you want to work on one, add a comment to the issue stating your interest and we can take another look at it.

Having said all of that, one of the best and most effective ways to contribute is to simply use the product in the scenarios that matter you the most and let us know if you find any issues.

ralmsdeveloper:
Thanks @Elion

PR #647, opened 2016-04-22
Added console write statements to improve user experience

### LINKED ISSUE #642: Show test progress when running functional tests
Currently when functional tests run there is not much output shown on the screen giving the users the impression that the tests are in hung state and users tend to kill the process.

I agree that the current experience is better compared to earlier one where we used to write bunch of log output statements and was not user friendly. Currently we show the detailed output only when the test fails which is great.

We should see if we can balance this experience so that we write minimal output to show the progress of tests and still show the detailed info only when the tests fail.

cc @natemcmaster

kichalla:
[Fixes #642] Show test progress when running functional tests

@Eilon @natemcmaster


Eilon:
:shipit:

PR #651, opened 2016-04-26
Split MusicStore into 10 projects

Eilon:
@DamianEdwards @barrytang @SajayAntony @davidfowl 

Hi folks,

I took MusicStore and split it into 10 projects (somewhat arbitrarily, though there's some logic to it).

We can use this project for some of the perf testing work in VS and from command line because this more closely reflects a small/mid-size solution. There is only minor code change to support the multi-project system (using MVC's new Application Parts feature). 99% of the change was just moving stuff around.

FYI: I'm not going to merge this PR, it's just for sharing. You can get everything from my server branch.

Thanks,
Eilon


Eilon:
My beautiful PR 😦

Issue #627, opened 2016-04-05
Re-enable tests skipped due to CoreFX bug


JunTaoLuo:
Revert the changes in https://github.com/aspnet/MusicStore/commit/a3b959a02837fe413987cd513a8d2439d42f05a7 once we have the CoreFX fix for https://github.com/dotnet/corefx/issues/7440

BrennanConroy:
This can be reverted now

JunTaoLuo:
Had to redisable these tests in https://github.com/aspnet/MusicStore/pull/659. Reopening this issue to track enabling of these tests after we have an updated cli.

Eilon:
@JunTaoLuo perhaps we can verify this again today? We should have fairly recent CoreFX's already, and today we've taken an even newer one.

JunTaoLuo:
@Eilon already doing that. There seems to be another problem, I'm verifying whether that is an issue with CoreFX.

Eilon:
Thanks!

JunTaoLuo:
blocked on https://github.com/dotnet/corefx/issues/8941

stephentoub:
@JunTaoLuo, can you try https://github.com/dotnet/corefx/pull/8946?

JunTaoLuo:
@stephentoub How do I test out your fix in the linked PR? Will you build a package for me to test?

stephentoub:
> How do I test out your fix in the linked PR? 

You can build System.Net.Http.dll from my branch and overwrite the one in your published directory.  If you don't want to build it yourself, I've emailed you a build you can use.

JunTaoLuo:
@stephentoub Thanks, I verified this has fixed the issue we are running into.

JunTaoLuo:
tests enabled after we received the fixes in https://github.com/dotnet/corefx/pull/9282

PR #659, opened 2016-05-04
Use in memory store for non-windows tests

JunTaoLuo:
I found this while debugging failures on non-windows CoreCLR build failure. These tests will still fail due to the content length issue but this needs to be fixed too.

cc @pranavkm

### DIFF
--- a/shared/Mocks/StartupOpenIdConnectTesting.cs
+++ b/shared/Mocks/StartupOpenIdConnectTesting.cs
@@ -42,11 +42,8 @@ namespace MusicStore
         {
             services.Configure<AppSettings>(Configuration.GetSection("AppSettings"));
 
-            //Sql client not available on mono
-            var useInMemoryStore = _platform.IsRunningOnMono;
-
             // Add EF services to the services container
-            if (useInMemoryStore)
+            if (_platform.UseInMemoryStore)
             {
                 services.AddDbContext<MusicStoreContext>(options =>
                             options.UseInMemoryDatabase());
--- a/src/MusicStore/Platform.cs
+++ b/src/MusicStore/Platform.cs
@@ -86,5 +86,14 @@ namespace MusicStore
                 return _isNano.Value;
             }
         }
+
+        // Sql client not available on mono, non-windows, or nano
+        public bool UseInMemoryStore
+        {
+            get
+            {
+                return !IsRunningOnWindows || IsRunningOnMono || IsRunningOnNanoServer;
+            }
+        }
     }
 }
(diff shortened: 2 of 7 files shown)


shared/Mocks/StartupOpenIdConnectTesting.cs
```diff
@@ -42,8 +42,10 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.Configure<AppSettings>(Configuration.GetSection("AppSettings"));
 
-            //Sql client not available on mono
-            var useInMemoryStore = _platform.IsRunningOnMono;
+            //Sql client not available on mono, non-windows, or nano
+            var useInMemoryStore = !_platform.IsRunningOnWindows
```
pranavkm: Can we just copy this property into `Platform` instead of repeating it multiple times over?

pranavkm:
:shipit:

JunTaoLuo:
:up::date: Will reopen https://github.com/aspnet/MusicStore/issues/627 to re-enable the tests again after we have new cli

src/MusicStore/Platform.cs
```diff
@@ -82,5 +82,14 @@ public bool IsRunningOnNanoServer
                 return _isNano.Value;
             }
         }
+
+        //Sql client not available on mono, non-windows, or nano
+        public bool UseInMemoryStore
```
victorhurdugaci: This is the wrong place for this property. Platform has nothing to do with stores

JunTaoLuo: We wanted to reduce the duplication of code, that's why we moved it.

victorhurdugaci: Reducing duplication is fine. All I'm saying is that this property should not be in this class

JunTaoLuo: I think it's reasonable for the platform class to provide indication whether SQL is available. Maybe I should rename this?

pranavkm: Seems ok to me. Maybe @kichalla has suggestions since he's pretty familiar with this piece of code.

src/MusicStore/Platform.cs
```diff
@@ -82,5 +82,14 @@ public bool IsRunningOnNanoServer
                 return _isNano.Value;
             }
         }
+
+        //Sql client not available on mono, non-windows, or nano
```
pranavkm: formatting - space after `//`

pranavkm:
@kichalla could you have a look? :shipit: from me.

kichalla:
:shipit:

JunTaoLuo:
Will merge, tested locally on windows and mac.

PR #662, opened 2016-05-09
Fixes null reference exception

DavidObando:
@sivagms @martinpf 

This dramatically reduces the content we're throwing into the log.

### DIFF
--- a/src/MusicStore/Controllers/ShoppingCartController.cs
+++ b/src/MusicStore/Controllers/ShoppingCartController.cs
@@ -71,19 +71,29 @@ namespace MusicStore.Controllers
                 .Include(c => c.Album)
                 .SingleOrDefaultAsync();
 
-            // Remove from cart
-            int itemCount = cart.RemoveFromCart(id);
+            string message;
+            int itemCount;
+            if (cartItem != null)
+            {
+                // Remove from cart
+                itemCount = cart.RemoveFromCart(id);
 
-            await DbContext.SaveChangesAsync(requestAborted);
+                await DbContext.SaveChangesAsync(requestAborted);
 
-            string removed = (itemCount > 0) ? " 1 copy of " : string.Empty;
+                string removed = (itemCount > 0) ? " 1 copy of " : string.Empty;
+                message = removed + cartItem.Album.Title + " has been removed from your shopping cart.";
+            }
+            else
+            {
+                itemCount = 0;
+                message = "Could not find this item, nothing has been removed from your shopping cart.";
+            }
 
             // Display the confirmation message
 
             var results = new ShoppingCartRemoveViewModel
             {
-                Message = removed + cartItem.Album.Title +
-                    " has been removed from your shopping cart.",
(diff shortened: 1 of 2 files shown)


src/MusicStore/Controllers/ShoppingCartController.cs
```diff
@@ -71,24 +71,39 @@ public ShoppingCartController(MusicStoreContext dbContext)
...
-
-            await DbContext.SaveChangesAsync(requestAborted);
-
-            string removed = (itemCount > 0) ? " 1 copy of " : string.Empty;
-
-            // Display the confirmation message
-
-            var results = new ShoppingCartRemoveViewModel
+            ShoppingCartRemoveViewModel results;
+            if(cartItem != null)
```
Eilon: Add space after `if`.

src/MusicStore/Controllers/ShoppingCartController.cs
```diff
@@ -71,24 +71,39 @@ public ShoppingCartController(MusicStoreContext dbContext)
...
+                    CartCount = await cart.GetCount(),
+                    ItemCount = itemCount,
+                    DeleteId = id
+                };
+            }
+            else
+            {
+                results = new ShoppingCartRemoveViewModel
+                {
+                    Message = "Unable to find the specified item from the cart",
```
Eilon: Should this be `Unable to *remove* the...`?

Eilon: And perhaps remove the word `specified` (nothing was "specified" from the user's perspective).

src/MusicStore/Controllers/ShoppingCartController.cs
```diff
@@ -71,24 +71,39 @@ public ShoppingCartController(MusicStoreContext dbContext)
...
+                // Remove from cart
+                int itemCount = cart.RemoveFromCart(id);
+
+                await DbContext.SaveChangesAsync(requestAborted);
+
+                string removed = (itemCount > 0) ? " 1 copy of " : string.Empty;
+
+                // Display the confirmation message
+
+                results = new ShoppingCartRemoveViewModel
```
anpete: Consider using variables for Message and ItemCount so that the view model creation is not duplicated.

anpete:
:shipit:

src/MusicStore/Controllers/ShoppingCartController.cs
DavidObando: Will do.

src/MusicStore/Controllers/ShoppingCartController.cs
DavidObando: Thanks, didn't put much thought into this. Will correct to something more appropriate.

src/MusicStore/Controllers/ShoppingCartController.cs
DavidObando: Will do.

PR #687, opened 2016-06-22
Changed CartItem.CartId type from string to Guid.  Needed this change…

martinpf:
Changed CartItem.CartId type from string to Guid.  Needed this change to ease the creation of an index on the CartId column in the database.  String is not an indexable type, guid (uniqueidentifier in SQL Server) is.  Adding that index has a big impact on database performance for the MusicStore app for stress and reliability testing


martinpf:
decided to commit to another branch

PR #688, opened 2016-06-22
Added centos and rhel rids to project.json

kichalla:
@BrennanConroy


BrennanConroy:
Should we add the other major supported OS's for when we create agents for them?

kichalla:
> Should we add the other major supported OS's for when we create agents for them?

Would be good to do it, but I would like to get this in soon so that I can verify this script in CI.

BrennanConroy:
:shipit:

PR #690, opened 2016-06-28
Fixed incorrect DNX targets

IEvangelist:
Wrapped `IWebHost` in a using for the `.Run` invocation. Fixed incorrect `.xproj` **DNX** targets.

### DIFF
--- a/test/E2ETests/E2ETests.xproj
+++ b/test/E2ETests/E2ETests.xproj
@@ -16,5 +16,5 @@
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
-  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.targets" Condition="'$(VSToolsPath)' != ''" />
+  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.targets" Condition="'$(VSToolsPath)' != ''" />
 </Project>
--- a/test/MusicStore.Test/MusicStore.Test.xproj
+++ b/test/MusicStore.Test/MusicStore.Test.xproj
@@ -16,5 +16,5 @@
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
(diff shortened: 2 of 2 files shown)


src/MusicStore.Standalone/Program.cs
```diff
@@ -44,9 +44,10 @@ public static void Main(string[] args)
                 builder.UseKestrel();
             }
 
-            var host = builder.Build();
-
-            host.Run();
+            using (var host = builder.Build())
```
Eilon: This doesn't seem to be the pattern that's used in all the project templates. The templates all just call `.Build()` and then `.Run()` without disposal:
* https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/WebAPI/Program.cs#L20-L22
* https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/EmptyWeb/Program.cs#L19-L21
* https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/StarterWeb/Program.cs#L19-L21

I would prefer keeping this to match what the templates do as I don't think there's any need to dispose anything here (the app will exit and clean up all resources after the last line anyway).

Eilon:
@IEvangelist thanks for the PR! Please see the comment I just left, plus the CLA bot is requesting that you sign the CLA, which is required for us to accept contributions.

src/MusicStore.Standalone/Program.cs
IEvangelist: Ironically, I am in the process of creating a pull request to address those specifically. Those templates should all be utilizing the `using` statement.

BrennanConroy: Like @Eilon said after `host.Run()` returns the app immediately exits which will automatically dispose the host. I prefer the current way we have it since it looks cleaner.

IEvangelist:
Hi @Eilon, 

I removed the change for the wrapping of the `IWebHost` instance in a `using`.  On the [`aspnet/Templates` I had an open issue](https://github.com/aspnet/Templates/pull/639) to make this more consistent - but after further discussion it was closed. I have also signed the CLA.

Thank you

Eilon:
@IEvangelist thanks!

@kichalla , all yours!

kichalla:
e0082643168de12b680f98294667bf8c291da416

Thanks! @IEvangelist

Issue #669, opened 2016-05-17
Cache of "topselling" items should be invalidated when an album is removed


mikeharder:
## Repro Steps
1. Launch MusicStore
2. Login as admin
3. Remove topselling album, such as `The Best Of The Men At Work`
4. Refresh homepage
## Expected

Removed album does not appear on homepage
## Actual

Removed album appears on homepage, but clicking it causes an `Item not found` error.

mkArtakMSFT:
Closing this issue as there was no community involvement for quite a while now.

PR #697, opened 2016-08-19
Fix for #669 

KostaVlev:
Fix for #669  "Cache of "topselling" items should be invalidated when an album is removed".


natemcmaster:
Thanks for the contribution. Sorry this one fell off the radar. I'm closing as at this point the changes are stale and wouldn't work. However, if you would like to take a stab at this one again, can you comment on https://github.com/aspnet/MusicStore/issues/669? That will let us know some one is "claiming" the up-for-grabs issue.

PR #702, opened 2016-09-04
Increase .travis.yml consistency between repos

dougbu:
- aspnet/Universe#349
- minimize `dotnet` setup time; no need for caching
- build with `--quiet`
- `KOREBUILD_TEST_DNXCORE` env variable isn't used anymore
- no need for libuv v1.4.2 w/ current Kestrel packages
- no need for icu4c w/ current `dotnet` builds


dougbu:
3ad40ba

PR #703, opened 2016-09-08
Increase HttpClient timeout

BrennanConroy:
@Tratcher @kichalla 

Win7-Universe tests fail a lot from what looks like a timeout, raising to 15 looks like it works on the failing machine


Tratcher:
:shipit:

PR #713, opened 2016-10-09
Add details on running in Docker Windows containers

friism:
This is based on work by @PatrickLang done here: https://github.com/PatrickLang/MusicStore/tree/plang-temp-compose

### DIFF
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -1,4 +1,4 @@
-FROM microsoft/dotnet:1.0.0-preview2-windowsservercore-sdk
+FROM microsoft/dotnet:1.0.0-preview2.1-nanoserver-sdk
 
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop';"]
 
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -1,0 +1,20 @@
+FROM microsoft/dotnet:1.0.0-preview2-windowsservercore-sdk
+
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop';"]
+
+# REMARK: Temporary workaround for Windows DNS client weirdness
+RUN set-itemproperty -path 'HKLM:\SYSTEM\CurrentControlSet\Services\Dnscache\Parameters' -Name ServerPriorityTimeLimit -Value 0 -Type DWord
+
+RUN New-Item -Path \MusicStore\samples\MusicStore.Standalone -Type Directory
+WORKDIR MusicStore
+
+ADD samples/MusicStore.Standalone/project.json samples/MusicStore.Standalone/project.json
+ADD NuGet.config .
+RUN dotnet restore --no-cache .\samples\MusicStore.Standalone
+
+ADD samples samples
+RUN dotnet build .\samples\MusicStore.Standalone
+
+EXPOSE 5000
+ENV ASPNETCORE_URLS http://0.0.0.0:5000
+CMD dotnet run -p .\samples\MusicStore.Standalone
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -10,7 +10,7 @@ WORKDIR MusicStore
 
 ADD samples/MusicStore.Standalone/project.json samples/MusicStore.Standalone/project.json
 ADD NuGet.config .
-RUN dotnet restore --no-cache .\samples\MusicStore.Standalone
+RUN dotnet restore .\samples\MusicStore.Standalone
 
 ADD samples samples
 RUN dotnet build .\samples\MusicStore.Standalone
--- a/docker-compose.windows.yml
+++ b/docker-compose.windows.yml
@@ -5,7 +5,7 @@ services:
     environment:
       sa_password: "Password1"
     ports:
-      - "1433:1433" # for debug. Remove this for production
+      - "1433:1433" # REMARK: This is currently required, needs investigation
 
   web:
     build:
--- a/docker-compose.windows.yml
+++ b/docker-compose.windows.yml
@@ -1,0 +1,24 @@
+version: '2'
+services:
+  db:
+    image: microsoft/mssql-server-2016-express-windows
+    environment:
+      sa_password: "Password1"
+    ports:
+      - "1433:1433" # for debug. Remove this for production
+
+  web:
+    build:
+      context: .
+      dockerfile: Dockerfile.windows
+    environment:
+      - "Data:DefaultConnection:ConnectionString=Server=db,1433;Database=MusicStore;User Id=sa;Password=Password1;MultipleActiveResultSets=True"
+    depends_on:
+      - "db"
+    ports:
+      - "5000:5000"
+
+networks:
+  default:
+    external:
+      name: nat
--- a/README.md
+++ b/README.md
@@ -25,6 +25,13 @@ This project is part of ASP.NET Core. You can find samples, documentation and ge
 * **[CustomHost]:**
 	6. Run `dnx . run` (This hosts the app in a console application - Application started at URL **http://localhost:5003/**).
 
+## Run on Docker Windows Containers
+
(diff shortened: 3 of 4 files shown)


rs38:
could it be modified to work with 
`FROM microsoft/dotnet:1.0.0-preview2-nanoserver-sdk`
?

friism:
@rs38 Yeah, that works great too. You have to run the windowsservercore-based sql server image though, so it doesn't matter greatly.

rs38:
isn't therey _any_ SQL based on nanoserver+.NET core image?

friism:
>isn't therey any SQL based on nanoserver
@rs38 Microsoft SQL Server does not run on Nano Server, as far as I know.

rs38:
that's why I mentioned "any".  Any in-process SQL would do, as well as not being bound to IIS but may be Kestrel. I need to admit I am not quite sure which prereq. EF has realating to SQL

friism:
@rs38 if you interested in a pure nanoserver-based sample, can you open a separate PR doing that? There's a list of EF supported databases here: https://docs.efproject.net/en/latest/providers/index.html

rs38:
I am afraid, I do not have a solution to provide

PatrickLang:
@rs38 MusicStore does have an in-memory database. I forgot exactly where, but you can hack up the code to enable it. I did that before I had the SQL connection working in containers.

docker-compose.windows.yml
```diff
@@ -0,0 +1,24 @@
...
+      sa_password: "Password1"
+    ports:
+      - "1433:1433" # REMARK: This is currently required, needs investigation
+
+  web:
+    build:
+      context: .
+      dockerfile: Dockerfile.windows
+    environment:
+      - "Data:DefaultConnection:ConnectionString=Server=db,1433;Database=MusicStore;User Id=sa;Password=Password1;MultipleActiveResultSets=True"
```
PatrickLang: does `db,1433` work? I have been doing `db:1433`

friism: Comma is the canonical representation, and just want to echo this commenter's sentiment: http://stackoverflow.com/questions/5294721/how-to-specify-a-port-number-in-sql-server-connection-string#comment27379349_5294741

PatrickLang:
Thanks for doing this, I've been intending to move [my version](https://github.com/PatrickLang/MusicStore/tree/plang-temp-compose/src/MusicStore) back to Kestrel and rebase it  and issue a PR but this looks cleaner

friism:
@PatrickLang totally! And this was based on your prepatory work for Ignite btw, adding acknowledgement in the description

friism:
@pranavkm @Praburaj I already signed the CLA last year, do I need to sign it again?

pranavkm:
@troydai do you think this would be a useful scenario for our functionals?

If not, perhaps these instructions might be better suited for the https://github.com/aspnet/Home repo. @danroth27 for thoughts on this.

friism:
> Do you have a chance to tryout the Music Store on Windows Nano Server?

@troydai as I mentioned [above](https://github.com/aspnet/MusicStore/pull/713#issuecomment-253850447), Windows Server core base layer is needed anyway for the SQL Server service. The web part runs fine on nano, and I can change it, but it means that for a user starting on a naked machine, they have to pull _both_ nanoserver and windowsservercore base layers

Dockerfile.windows
```diff
@@ -0,0 +1,20 @@
...
+ADD samples/MusicStore.Standalone/project.json samples/MusicStore.Standalone/project.json
+ADD NuGet.config .
+RUN dotnet restore --no-cache .\samples\MusicStore.Standalone
+
+ADD samples samples
+RUN dotnet build .\samples\MusicStore.Standalone
+
+EXPOSE 5000
+ENV ASPNETCORE_URLS http://0.0.0.0:5000
+CMD dotnet run -p .\samples\MusicStore.Standalone
```
friism: @troydai yes, but start-up is faster this way. It seems more appropriate to have compilation done during build than during run. It's also nice to know that the code actually compiles, before trying to run it.

Dockerfile.windows
```diff
@@ -0,0 +1,20 @@
...
+
+# REMARK: Temporary workaround for Windows DNS client weirdness
+RUN set-itemproperty -path 'HKLM:\SYSTEM\CurrentControlSet\Services\Dnscache\Parameters' -Name ServerPriorityTimeLimit -Value 0 -Type DWord
+
+RUN New-Item -Path \MusicStore\samples\MusicStore.Standalone -Type Directory
+WORKDIR MusicStore
+
+ADD samples/MusicStore.Standalone/project.json samples/MusicStore.Standalone/project.json
+ADD NuGet.config .
+RUN dotnet restore --no-cache .\samples\MusicStore.Standalone
```
friism: `--no-cache` is not about using the cache already in the image, at least that's not how I understand the flag description:

    Do not cache packages and http requests.

`--no-cache` means that `dotnet` won't cache the packages restored. That's desirable here, since the cache would not serve any purpose, and would just crud up the image.

rs38:
Nano image is less than a 10th of servercore. Who would hesitate? ☺

Dockerfile.windows
troydai: `--no-cache` actually means don't use the cached http response from last restore and hit the http endpoint directly to ensure the latest packages are pulled. It doesn't mean packages are not stashed in `~/.nuget` folder. You can do a little experiment here. Delete the `~/.nuget` folder and run `dotnet restore --no-cache`. The packages are downloaded regardless.

In order to save the disk, you need to remove the `~/.nuget` as well as the `%LocalAppData%/NuGet/Cache` folder.

I agree the description is misleading here.

friism:
@rs38 one would hesitate because in this specific case, switching to Nanoserver causes Docker to download _more_ stuff on an empty system (because windowsservercore is required no matter what for SQL Server)

Dockerfile.windows
troydai: I see what you mean. Then the better pattern here will be:

```
# publish the application
RUN dotnet publish -o <output folder> -c Release

# initial command
CMD dotnet <output folder>\<assembly>
```

friism: @troydai I don't think doing a full publish is necessary, and it takes longer. Simply running `dotnet build` and `dotnet run` is more akin to what one would with simple iterative non-container development. It also makes it easier to use this container if one wants to switch between Debug and Release.

friism:
@troydai ptal

friism:
Alright, @taylorb-microsoft convinced me to change to nanoserver because it starts and builds faster. Also pulled in some compose improvements from Patrick. For now, this requires master builds of compose: https://dl.bintray.com/docker-compose/master/

rs38:
this cannot run on nanoserver (without hyper-v like as e.g. on Azure) as long as there is a dependency to a windowsservercore based SQL db.

Dockerfile.windows
rs38: just tried to build with docker build (instead of compose), did I miss sth.?:   
(edit due to incorrect paste)
`Step 8/13 : RUN dotnet restore .\samples\MusicStore.Standalone
 ---> Running in 90630e2fe7d6
log  : Restoring packages for C:\MusicStore\samples\MusicStore.Standalone\project.json...
error: Unable to load the service index for source https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json.
error:   An error occurred while sending the request.
error:   The server name or address could not be resolved.`

rs38: I apologize. Due to some virt NIC changes there was no net connection from inside the container. I found the answer by yourself  in another repo issue  :)

PatrickLang:
Are there any objections to merging this if the AppVeyor build is fixed?

friism:
It looks like the Appveyor build has been broken for approximately 5 months...

glennc:
@muratg Are the guys responsible for build breaks not looking at these?

PR #717, opened 2016-10-24
Use AspNetCore meta package

BrennanConroy:
@Eilon @muratg


Eilon:
@DamianEdwards @muratg @glennc  should we keep this or actually undo it given some of the other ongoing discussions we're having?

muratg:
I unlisted this package.  But for now, we should mark it as "no-ship" on packages.csv file

PR #719, opened 2016-10-26
Fix NuGet.Config name casing

mfilippov:
If NuGet.config has an incorrect case, dotnet restore don't use it on a case-sensitive fs.


Eilon:
Hmm, `NuGet.config` is the case that's used everywhere as far as I can tell. Are you sure this is the wrong case here?

Eilon:
@yishaigalatzer ?

mfilippov:
@Eilon This commit helps me restore packages on mac with case-sensitive fs.

Eilon:
@mfilippov I understand the purpose, I'm just not sure that it is wrong. We build all these repos on Windows, Mac, and Linux, and haven't had any issues.

emgarten:
NuGet treats OSX as case insensitive and does not check all casings of the file name. I've opened https://github.com/NuGet/Home/issues/4019 to track the issue. 

Ideally the ``NuGet.Config`` casing should be used every. I recommend taking this PR.

Eilon:
@emgarten how do our Ubuntu builds work, then? Does NuGet still allow `NuGet.config` on case-sensitive file systems? Or is it that we should use the "most preferred" name to avoid the fallbacks?

emgarten:
@Eilon on Ubuntu NuGet will find ``NuGet.config``. For perf reasons NuGet skips the extra file casing checks on OSes that it expects to be case-insensitive, when it does that it checks only for ``NuGet.Config``. There isn't a good way to tell from a managed app if the file system is actually case sensitive or not, but this could be fixed just by always checking for the extra casings on OSX in NuGet.

natemcmaster:
@mfilippov thanks for the contribution. Sorry this fell off our radar for a while. At this point, the PR is unmergeable because there are file conflicts. I'm closing for now as it appears the underlying issue is already tracked https://github.com/NuGet/Home/issues/4019

PR #721, opened 2016-10-28
Use WindowsIdentity instead of environment variables to verify NTLM a…

pranavkm:
…uth scenarios

This is required to fix tests when tests are run under a system account.


pranavkm:
cc @kichalla \ @Tratcher

Issue #710, opened 2016-10-06
Please add PostgreSQL support


JanSichula:
May I kindly suggest that you add support for PostgreSQL. This way, this referential solution will be truly cross-platform and not dependent directly on MS SQL and indirectly on Windows. There are folks out there who are excited about ASP.NET Core and yet would like to avoid any lock-in into closed source proprietary databases.

Eilon:
See explanation here: https://github.com/aspnet/MusicStore/pull/730#issuecomment-267112052

PR #730, opened 2016-12-01
Support using PostgreSQL as a database (via Npgsql)

tmds:


tmds:
Fixes https://github.com/aspnet/MusicStore/issues/710

tmds:
Can I get some feedback on the PR?

Hussain-arif:
feedback of what type ??????

tmds:
> feedback of what type ??????
Code review to get the PR merged.

tmds:
@Tratcher can you or someone else have a look please?

Eilon:
@tmds thanks for sending the PR, however for the time being the MusicStore app here is really just a test app that we use for our end-to-end test automation. As such, we generally don't add anything that won't be used by our automated tests because it increases the cost of maintaining this app.

If you'd like to maintain a separate copy of MusicStore in your own repo that is modified to show alternative options, such as using PostgreSQL, we would be happy to have a link from the readme of this repo to point to that.

PR #732, opened 2016-12-21
Updated content root of musicstore.standalone to enable debugging in VS

kichalla:
@Eilon all the tests pass with this change

Example of how the output directory looks like:

image

### DIFF
--- a/samples/MusicStore.Standalone/Program.cs
+++ b/samples/MusicStore.Standalone/Program.cs
@@ -1,4 +1,6 @@
 using System;
+using System.IO;
+using System.Reflection;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Net.Http.Server;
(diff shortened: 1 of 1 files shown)


samples/MusicStore.Standalone/Program.cs
```diff
@@ -14,10 +16,13 @@ public static void Main(string[] args)
...
                 .Build();
 
+            var currentExecutingAssemblyFileInfo = new FileInfo(typeof(Program).GetTypeInfo().Assembly.Location);
+
             var builder = new WebHostBuilder()
                 .UseConfiguration(config)
                 .UseIISIntegration()
-                .UseStartup("MusicStore.Standalone");
+                .UseStartup("MusicStore.Standalone")
+                .UseContentRoot(currentExecutingAssemblyFileInfo.Directory.FullName);
```
kichalla: To prevent duplicating logic from musicstore project, we copy the content files like Views, config files to the output folder of the standalone project and because of this we would have to change the location of the content root here.

PR #745, opened 2017-02-16
fork RemoveFromCart

vmural:


Eilon:
Hi @vmural can you share more info about this PR?

kimbaudi:
The "Remove From Cart" UI doesn't work and @vmural is trying to fix it. Currently, the RemoveFromCart method in ShoppingCartController returns a ShoppingCartRemoveViewModel as a JsonResult that has properties which are PascalCase (i.e., Message, CartTotal, CartCount, ItemCount, DeleteId). However, the jQuery ajax call gets the data as camelCase (i.e., message, cartTotal, cartCount, itemCount, deleteId). As a result, removing an item from the cart isn't working correctly. 

Looking at @vmural's commit, shows that he is missing '}' which is causing an even bigger error.

I'll try to fork this repo and commit the fix.

Eilon:
Closing this; we'll look at https://github.com/aspnet/MusicStore/pull/762 instead. Thanks!

PR #754, opened 2017-03-08
Fix NanoServer tests

kichalla:


test/E2ETests/SmokeTestsOnNanoServer.cs
```diff
@@ -263,6 +264,12 @@ public SmokeTestsOnNanoServer(ITestOutputHelper output, RemoteDeploymentConfig c
                     ApplicationBaseUriHint = applicationBaseUrl,
                     ApplicationType = applicationType
                 };
+
+                if (applicationType == ApplicationType.Standalone)
+                {
+                    deploymentParameters.AdditionalPublishParameters = "-r win10-x64";
```
kichalla: I am hardcoding this because NanoServer's RID is for Windows 2016 but our CI servers seem to be either Win 2012 or less (which have a different RID)...NanoServer tests work by publishing to a temp folder on the CI server and then copy bits to the target NanoServer

natemcmaster: Might be good to add an assert, just in case this code is every copy+pasted into another place.

samples/MusicStore/Program.cs
```diff
@@ -1,8 +1,9 @@
 using System;
-using System.IO;
+using System.Reflection;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Server.HttpSys;
 using Microsoft.Extensions.Configuration;
+using System.IO;
```
natemcmaster: Nit: sort

samples/MusicStore/Program.cs
```diff
@@ -15,8 +16,10 @@ public static void Main(string[] args)
                 .AddEnvironmentVariables(prefix: "ASPNETCORE_")
                 .Build();
 
+            var currentAssemblyLocation = new FileInfo(typeof(Program).GetTypeInfo().Assembly.Location);
+
             var builder = new WebHostBuilder()
-                .UseContentRoot(Directory.GetCurrentDirectory())
+                .UseContentRoot(currentAssemblyLocation.Directory.FullName)
```
natemcmaster: This isn't how we're supposed to use WebHostBuilder. It should be set to Directory.GetCurrentDirectory so that VS works correctly. What problem are you trying to resolve here?

test/E2ETests/SmokeTestsOnNanoServer.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.Extensions.Logging;
 using Xunit;
 using Xunit.Abstractions;
+using Microsoft.DotNet.PlatformAbstractions;
```
natemcmaster: Nit: sort

samples/MusicStore/Program.cs
kichalla: True this breaks VS experience. When the NanoServer tests get executed they currently do not run from the same directory as where the published output is (another reason why GetCurrentDirectory usage is not always correct). I will try fixing NanoServer tests instead. 

BTW, we used to have `copyToOutput` in project.json world, do we have something equivalent in msbuild build too?

natemcmaster: > When the NanoServer tests get executed they currently do not run from the same directory as where the published output

Let's change the tests and not break VS. People use this for demos an stuff.

Examples:
```xml
<ItemGroup>
  <None Include="notes.txt" CopyToOutputDirectory="Always" />
  <!-- CopyToOutputDirectory = { Always, PreserveNewest, Never } -->

  <Content Include="files\**\*" CopyToPublishDirectory="PreserveNewest" />
  <None Include="publishnotes.txt" CopyToOutputDirectory="Always"  CopyToPublishDirectory="Always" />
  <!-- CopyToPublishDirectory = { Always, PreserveNewest, Never } -->
</ItemGroup>
```
more details: http://www.natemcmaster.com/blog/2017/01/19/project-json-to-csproj/#files

PR #762, opened 2017-03-18
fix remove from cart ajax done

kimbaudi:
This pull request fixes an issue with RemoveFromCart. Currently, the RemoveFromCart method in ShoppingCartController returns a ShoppingCartRemoveViewModel as a JsonResult that has properties which are PascalCase (i.e., Message, CartTotal, CartCount, ItemCount, DeleteId). However, the jQuery ajax call gets the data as camelCase (i.e., message, cartTotal, cartCount, itemCount, deleteId). As a result, removing an item from the cart isn't working correctly. Changing the JSON from PascalCase to camelCase fixes the problem.

### DIFF
--- a/samples/MusicStore/Views/ShoppingCart/Index.cshtml
+++ b/samples/MusicStore/Views/ShoppingCart/Index.cshtml
@@ -35,15 +35,15 @@
                     }).done(function (data) {
                         // Successful requests get here
                         // Update the page elements
-                        if (data.ItemCount == 0) {
-                            $('#row-' + data.DeleteId).fadeOut('slow');
+                        if (data.itemCount == 0) {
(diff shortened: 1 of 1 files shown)


Eilon:
@kimbaudi thanks for sending this updated PR. I wonder if this has been broken for a long time when MVC changed the default JSON serialization to use camelCase instead of PascalCase?

kimbaudi:
@Eilon I believe MVC recently started using camelCase by default since 1.1.0 (so not too long ago).

Eilon:
Actually it was changed right before 1.0.0 RTM: https://github.com/aspnet/Announcements/issues/194

Anyway, the change looks good, so I'll merge!

PR #765, opened 2017-03-22
fix cart count

kimbaudi:
The cart count is wrong when the same album is added to the cart multiple times.


kimbaudi:
I'm not sure why AppVeyor build is failing with my pull request. Seems to do with E2ETesting, which I haven't touched.

PR #783, opened 2017-04-25
Enable validate scopes in dependency injection

pakrym:


samples/MusicStore/Models/SampleData.cs
```diff
@@ -21,14 +21,15 @@ public static class SampleData
         {
             using (var serviceScope = serviceProvider.GetRequiredService<IServiceScopeFactory>().CreateScope())
```
davidfowl: Since you're here, fix this to use the new extension method `CreateScope` on `IServiceProvider`

PR #787, opened 2017-05-03
React to Logging API changes

BrennanConroy:


samples/MusicStore/Program.cs
```diff
@@ -47,6 +48,15 @@ public static void Main(string[] args)
                 builder.UseKestrel();
             }
 
+            builder.ConfigureLogging(factory =>
+            {
+                factory.AddConsole();
+
+                var logLevel = string.Equals(environment, "Development", StringComparison.Ordinal) ? LogLevel.Information : LogLevel.Warning;
+
+                factory.AddFilter("Console", level => level >= logLevel);
```
JunTaoLuo: We need to bring back `level => logLevel <= level`

Issue #800, opened 2017-06-19
Smoke test fails on machine with cs-CZ culture


svick:
When I run an E2E smoke test on my Windows machine with cs-CZ culture, the test fails.

Specifically, the error is (truncated):

> Test Name:	E2ETests.SmokeTests_X64.WindowsOS(serverType: Kestrel, architecture: x64, applicationType: Portable)
Test FullName:	E2ETests.SmokeTests_X64.WindowsOS (c54ac54dac7f1e92fff5efff88317a1ccbd90d2e)
Test Source:	 : line 0
Test Outcome:	Failed
Test Duration:	0:00:32,487
>
> Result StackTrace:	
at E2ETests.Validator.<CreateAlbum>d__31.MoveNext() in E:\Users\Svick\git\aspnet-MusicStore\test\MusicStore.E2ETests\Implementation\Validator.cs:line 400
…
Result Message:	
Assert.Equal() Failure
Expected: http\://localhost:6861/Admin/StoreManager
Actual:   http\://localhost:6861/Admin/StoreManager/create

The returned page contains the following validation errors:

> The field Price must be a number.
> The field Price must be between 0,01 and 100.
> The value &#x27;9.99&#x27; is not valid for Price.

This looks like an issue caused by attempting to use the current culture (cs-CZ), which uses `,` as the decimal separator, to parse the inputted value, which uses `.`.

PR #801, opened 2017-06-19
Use localization middleware

svick:
Fixes https://github.com/aspnet/MusicStore/issues/800 by forcing the culture to be `en-US` using the localization middleware.

For a non-localized application, I believe this is the correct behavior.

Also, are five different versions of `Startup.cs` really necessary? Couldn't the common parts be extracted to a shared file? I personally had issues with this, when the smoke test kept failing even after I fixed the three `Startup`s in the root directory.


samples/MusicStore/ForTesting/Mocks/StartupOpenIdConnectTesting.cs
```diff
@@ -113,6 +115,15 @@ public void ConfigureServices(IServiceCollection services)
 
         public void Configure(IApplicationBuilder app)
         {
+            var supportedCultures = new[] { new CultureInfo("en-US") };
+
+            app.UseRequestLocalization(new RequestLocalizationOptions
```
pranavkm: Could you add a comment stating why you're forcing `en-US`? Would be helpful for posterity.

svick: Done.

natemcmaster:
@svick to get CI tests to pass on Linux, you'll need to rebase on the latest changes in dev branch so your PR has this fix: https://github.com/aspnet/MusicStore/commit/0c7c05e1ce3f1d16b2caa605cf0c929bd6a24c6f

svick:
@natemcmaster Done.

pranavkm:
Thanks for the PR @svick

PR #811, opened 2017-06-30
Use new Auth API

HaoK:


samples/MusicStore/ForTesting/Mocks/StartupSocialTesting.cs
```diff
@@ -94,7 +94,7 @@ public void ConfigureServices(IServiceCollection services)
                 options.AddPolicy("ManageStore", new AuthorizationPolicyBuilder().RequireClaim("ManageStore", "Allowed").Build());
             });
 
-            services.AddFacebookAuthentication(options =>
+            services.AddAuthentication().AddFacebook(options =>
```
natemcmaster: Nit: we typically add a newline before each of the `.AddXXX` call that hang off builders like .AddAuthentication

Issue #794, opened 2017-05-29
Error docker-compose build 


takekazuomi:
"Docker-compose build" fails.

## Reproduction procedure
```
$ docker-compose -f .\docker-compose.windows.yml build
db uses an image, skipping
Building web
Step 1/12 : FROM microsoft/dotnet:1.1-sdk-msbuild
1.1-sdk-msbuild: Pulling from microsoft/dotnet
ERROR: Service 'web' failed to build: image operating system "linux" cannot be used on this platform
```

## Problem
1.  Dockerfile.windows use old image.  https://github.com/aspnet/MusicStore/issues/777
2.  Dockerfile.windows use dotnet sdk 1.1 but Music Store app use 2.0.

natemcmaster:
Closing as we have removed dockerfiles from this repo. See #812 for more details.

Issue #777, opened 2017-04-12
Dockerfile.windows uses a deprecated base image


devinvisible:
From https://blogs.msdn.microsoft.com/dotnet/2017/03/07/announcing-net-core-tools-1-0/#docker

> We adopted changes to the Docker tag scheme we use at microsoft/dotnet. The “-projectjson” and “-msbuild” differentiation is now gone. The tags are significantly simpler, as you can see from the tags above.

natemcmaster:
Closing as we have removed dockerfiles from this repo. See #812 for more details.

Issue #776, opened 2017-04-12
Powershell doesn't exist in the Dockerfile base image


devinvisible:
The SHELL instruction specifies powershell but the base image doesn't have it available. 
(Solution 1) Perhaps the intention was to use the `-nanoserver` flavor?

In fact, I wonder if powershell is really necessary here. It's used for one command (`RUN New-Item -Path \MusicStore\samples\MusicStore -Type Directory`) which isn't even actually necessary - the WORKDIR and ADD instructions both create directories if they don't already exist.
(Solution 2) Remove the RUN New-Item line

natemcmaster:
We've decided to remove the Dockerfiles from this repo as building MusicStore in Docker is not supported. See #812.

Issue #735, opened 2017-01-17
Compiling errors


easec:
Hi there! Get a lot of compiling errors when I tried these on Windows Server 2016 Datacenter with latest Docker and Docker-Compose. Running the server in nested virtualization.

![error](https://cloud.githubusercontent.com/assets/23214234/22023981/c6e5f0b6-dcc8-11e6-8c0b-b4e8cdc349a9.png)

Any suggestions?

natemcmaster:
Compiling this repo in Docker isn't supported. We've decided to remove the docker files. See #812.

PR #812, opened 2017-07-03
Remove Dockerfiles

natemcmaster:
The version of .NET Core that MusicStore builds with will almost always be newer than the version available in the `microsoft/aspnetcore-nightly` or `microsoft/dotnet-nightly` Docker  images. Because nothing on the CI uses these files, the end up horribly stale and broken, and it is very difficult to make dockerfiles work for this repo.

Accordingly, let's remove Docker from this repo and update the docs to make it clearer that this repository is meant for our testing purposes only, and is not a "product" application sample.

Closes https://github.com/aspnet/MusicStore/issues/735.
Closes https://github.com/aspnet/MusicStore/issues/776.
Closes https://github.com/aspnet/MusicStore/issues/794.
Closes https://github.com/aspnet/MusicStore/issues/777


Eilon:
👍

PR #818, opened 2017-07-21
2.0.0-rtm to 2.1.0-preview1

ryanbrandenburg:


pranavkm:
Nothing in this repo needs to be versioned. We should delete the file.

PR #821, opened 2017-08-18
Install store from artifact dependency

JunTaoLuo:
Instead of Build.RS package

Simulated locally and the test passed. Already updated CI configuration with the necessary installation steps.


JunTaoLuo:
The only problem is that you would need to install a runtime store before being able to run this locally.

test/MusicStore.E2ETests/SmokeTestsUsingStore/TestHelper.cs
```diff
@@ -49,7 +52,7 @@ public TestHelper(ITestOutputHelper output) : base(output)
                         DbUtils.CreateConnectionString(musicStoreDbName)));
 
                 deploymentParameters.EnvironmentVariables.Add(
-                    new KeyValuePair<string, string>("DOTNET_SHARED_STORE", storeDirectory));
+                    new KeyValuePair<string, string>("DOTNET_SHARED_STORE", Environment.GetEnvironmentVariable("DOTNET_SHARED_STORE")));
```
natemcmaster: Don't child processes automatically inherit all environment variables? Or maybe it's too late at night to remember this detail correctly.

JunTaoLuo: Yup this shouldn't be needed.

PR #822, opened 2017-08-25
React to patch version changes in 1.0.x

natemcmaster:


src/MusicStore.Standalone/project.json
```diff
@@ -3,7 +3,7 @@
     "Microsoft"
   ],
   "description": "Music store application on ASP.NET 5",
-  "version": "1.0.3",
```
JunTaoLuo: It's strange that we have these samples versioned but this is a relic of the past I guess.

PR #828, opened 2017-10-26
Add RuntimeStore config to music store

natemcmaster:
React to https://github.com/aspnet/MetaPackages/pull/244 - changing the order in which our repos build and test.


natemcmaster:
merging to unblock ci

Issue #830, opened 2017-11-13
Tests failing on CI for MusicStore


jkotalik:
```
System.IO.FileNotFoundException : AspNetCoreModule could not be found.
   at Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer.<StartIISExpressAsync>d__12.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer.<DeployAsync>d__11.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at E2ETests.NtlmAuthenticationTests.<NtlmAuthenticationTest>d__7.MoveNext() in C:\b\w\33bdfc1cae7b2a38\modules\MusicStore\test\MusicStore.E2ETests\NtlmAuthentationTest.cs:line 93
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
------- Stdout: -------
| TestLifetime Information: Starting test NtlmAuthentication_IISExpress_Clr_Standalone
| Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Information: Deploying [Variation] :: ServerType=IISExpress, Runtime=Clr, Arch=x64, BaseUrlHint=, Publish=True
[...]

PR #831, opened 2017-11-13
Disabling tests while investigating file location issues

jkotalik:
For #830


jkotalik:
Seems to be failures with Kestrel Tests here. I don't think they are related to my errors, should I rerun appveyor?

pakrym:
@jkotalik I think that's a known issue in Kestrel  https://github.com/aspnet/KestrelHttpServer/issues/2116

PR #838, opened 2018-01-05
Make the ANCM a conditional dependency 

jkotalik:
As ANCM is now an internal dependency, we can't build it on Mac or Linux.


samples/MusicStore/MusicStore.csproj
```diff
@@ -14,7 +14,7 @@
     <Content Update="ForTesting\**\*" CopyToPublishDirectory="Never" Condition=" '$(PublishForTesting)' != 'true' " />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(OS)' != 'Windows_NT'">
```
natemcmaster: isn't this backwards?

jkotalik: Totally is

PR #841, opened 2018-03-12
[Testing] Debugging some tests that only fail on High Sierra

Tratcher:
Caused by dotnet/corefx#27912 ?


Tratcher:
@ryanbrandenburg please review, this fixes the current test failures on High Sierra by forcing the HTTP version to 1.1.

PR #845, opened 2018-04-16
Use the Razor.Sdk package

pranavkm:
We made a breaking change to the Razor Sdk (https://github.com/aspnet/Razor/commit/95835d6c37c108ae0788f49ccc0d17f3ba20586e) that requires using a version of Microsoft.NET.Sdk.Razor newer than the one in the CLI. This resolves the issue and fixes test failures.


pranavkm:
Fixes https://github.com/aspnet/MusicStore/issues/846

PR #850, opened 2018-05-16
Move to TestMatrix approach

Tratcher:
In reaction to https://github.com/aspnet/Hosting/pull/1412

Mechanical and organizational changes only.

PR #851, opened 2018-05-16
Add missing config files

Tratcher:
Followup to https://github.com/aspnet/MusicStore/pull/850.

IIS Express tests are failing for OIDC and PublishAndRun. These tests didn't used to have IIS Express or nginx variations so they did not include the needed config files.


Tratcher:
cc @ryanbrandenburg RE: http://aspnetci/viewLog.html?buildId=469875&tab=buildResultsDiv&buildTypeId=Lite_MusicStoreTest

test/MusicStore.E2ETests/nginx.conf
```diff
@@ -0,0 +1,37 @@
+﻿error_log [errorlog];
```
pakrym: Is it copied from somewhere? Can we have a default version in Hosting?

Tratcher: Copied from server tests. Filed https://github.com/aspnet/Hosting/issues/1423

Tratcher:
Updated. Getting NGinx to work is more involved. I've removed it for now to unblock fixing the IIS tests.

Issue #876, opened 2018-11-06
Test failure: PublishAndRun_Test


aspnet-hello:
This issue was made automatically. If there is a problem contact ryanbrandenburg.

This test [failed](http://aspnetci/viewLog.html?buildId=588636&buildTypeId=XPlat_Windows_Win2008r2_Universe) with the following error:

```
Assert.Equal() Failure
Expected: OK
Actual:   BadGateway
   at E2ETests.Validator.VerifyHomePage(HttpResponseMessage response, Boolean useNtlmAuthentication) in C:\BuildAgent\work\e37dd45d8cd1eaf4\modules\MusicStore\test\MusicStore.E2ETests\Implementation\Validator.cs:line 94
   at E2ETests.PublishAndRunTests.PublishAndRun_Test(TestVariant variant) in C:\BuildAgent\work\e37dd45d8cd1eaf4\modules\MusicStore\test\MusicStore.E2ETests\PublishAndRunTests.cs:line 81
--- End of stack trace from previous location where exception was thrown ---
------- Stdout: -------
| [0.005s] TestLifetime Information: Starting test PublishAndRun_Test-Server: IISExpress, TFM: netcoreapp3.0, Type: Portable, Arch: x86, ANCM: V2, Host: OutOfProcess at 2018-11-06T11:11:48
| [0.001s] TestLifetime Information: Starting test PublishAndRunTests_Server: IISExpress, TFM: netcoreapp3.0, Type: Portable, Arch: x86, ANCM: V2, Host: OutOfProcess at 2018-11-06T11:11:48
| [0.002s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISExpressDeployer Information: Deploying [Variation] :: ServerType=IISExpress, Runtime=CoreClr, Arch=x86, BaseUrlHint=, Publish=True
| [0.002s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISExpressDeployer Information: Executing command dotnet publish  --output "C:\BuildAgent\temp\buildTmp\c752a86b27bf44d48ee0e55d55bd10b5" --framework netcoreapp3.0 --configuration Release --no-restore -p:VerifyMatchingImplicitPackageVersion=false 
| [0.254s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISExpressDeployer Information: dotnet-publish stdout: Microsoft (R) Build Engine version 15.9.19+g938f3292a0 for .NET Core
[...]

PR #877, opened 2018-11-07
Disable IISExpress tests

jkotalik:
Lingering tests that are failing as seen in #876 .


jkotalik:
Going to go ahead and merge as it’s just for failing tests.

PR #879, opened 2018-11-13
Switch async void to async Task

javiercn:
This repository is a test application used for ASP.NET Core internal test processes.
It is not intended to be a representative sample of how to use ASP.NET Core.

Samples and docs for ASP.NET Core can be found here: https://docs.asp.net.


natemcmaster:
> D:\a\1\s\test\MusicStore.Test\Models\ShoppingCartTest.cs(22,27): error CS1983: The return type of an async method must be void, Task or Task<T> [D:\a\1\s\test\MusicStore.Test\MusicStore.Test.csproj]

